PR #259, opened 2020-03-26
Web UI benchmark examples and changes to ease the UI implementation

tgy:


tgy:
@demianw should this be closed?

PR #513, opened 2020-10-12
add method to load non-reported activations

tgy:


tgy:
closing this as it was a workaround before negation was implemented

PR #522, opened 2020-10-23
Implementation of within language MARG queries

demianw:
The main idea is:

At the intermediate representation level:
a new probabilistic predicate e.g. 
`Implication(A(x, y), Condition(B(x, z), C(y, z))` which has the semantic that the probability of `A(x, y)` corresponds to `B(x, _) | C(y, _)`

At the frontend level, to be implemented, this could be with an expression
```
A[x, y, PROB(x, y)] = B(x, z) // C(y, z)
```

So far I implemented the representation of the marg query at the IR level with a simple test and adapted `neurolang.probabilistic.query_resolution.compute_probabilistic_solution`

@tgy, if you agree with this proposal, can you take over? You will need this to be finished for your subsequent work. It needs
* A bit more testing at the IR level
* Front-end expression (__floordiv__) operator in the case of a predicate which could be a new pattern on the TranslateProbabilisticQueryMixin
* Testing at the front-end level


tgy:
now awaiting #537

tgy:
now replaced by #538

PR #530, opened 2020-10-28
Handling of constants in query flattening

tgy:
-   shattering is not unifying equalities anymore. instead, equalities are kept
    within the query's formula and directly translated to relational algebra
    operations (notably extended projections).
-   extended projection implemented for very specific cases in the
    `ProbSemiringSolver` of the `dichotomy_theorem_based_solver` module. the
    only case for which the extended projection is performed without raising an
    exception is when:
    -   all non-provenance columns of the provenance set are kept the same
        during the projection using a `{c: c}` eval expression. this ensures
        that the resulting number of tuples remains the same and so the
        provenance column does not need to be updated. if it was not the case,
        then the operations to correctly update the provenance column to
        maintain correct probabilities would be undefined in many cases.
    -   the provenance column is not part of the projection list, although it
        is added during the actual projection to maintain the provenance column
        in the underlying relational algebra set
-   flattening was modified to leverage the MGU algorithm internally to
    maintain variable equalities within the conjunctive queries that can later
    be translated to relational algebra operations. internally, before applying
    the MGU algorithm, the conjunctive queries (rules of the program) are
    normalised in two ways
    -   first, any constant in the head predicate is replaced by a fresh symbol
        and a variable-to-constant equality is added to the body to maintain
        the rule's semantic
    -   second, any repetition of a variable in the head predicate is replaced
        by a fresh symbol and a between-variables equality is added to the body
        of the rule to maintain the rule's semantic


tgy:
replaced by #537

PR #538, opened 2020-11-04
Within-language MARG queries

tgy:


neurolang/frontend/probabilistic_frontend.py
```diff
@@ -75,8 +78,9 @@ class NeurolangPDL(QueryBuilderDatalog):
 
     def __init__(
         self,
-        chase_class: Type[Chase] = Chase,
+        chase_class=Chase,
```
demianw, line 81: Why are you removing the type annotations? Is it on purpose?

demianw (commented):
Just small comments

neurolang/frontend/probabilistic_frontend.py
tgy: oops, this was accidentally removed during a merge conflict resolution. i just added it back.

PR #548, opened 2020-11-09
Query-based probabilistic facts

tgy:


tgy:
This docstring describes my approach

    Translate a query-based probabilistic fact to two rules. A deterministic
    rule that takes care of inferring the set of tuples in the probabilistic
    table and a probabilistic rule that transforms the result from the
    deterministic resolution into a proper probabilistic table by re-attaching
    the deterministically-obtained probabilities.
    For example, the rule `P(x) : f(x) :- Q(x)` is translated into the rules
        _f1_(_f2_, x) :- Q(x), _f2_ = f(x)
        P(x) : _f2_ :- _f1_(_f2_, x)
    where `_f1_` and `_f2_` are fresh symbols. Note: we make sure not to
    expose the resulting `_f1_` relation to the user, as it is not part of
    its program.

@demianw I would like to know if you think this is the right approach

neurolang/expressions.py
```diff
@@ -863,3 +863,10 @@ def pop_scope(self):
         if es is None:
             raise NeuroLangException('No enclosing scope')
         self.symbol_table = self.symbol_table.enclosing_scope
+
+
+MATMUL = Constant(
```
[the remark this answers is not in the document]
demianw: I think that this represents non-consistent semantics. Either all operators are here or none. For now, let's go for none.

neurolang/probabilistic/expressions.py
```diff
@@ -24,6 +24,14 @@ def __init__(self, probability, body):
         self.body = body
         self._symbols = body._symbols | self.probability._symbols
 
+    @property
+    def functor(self):
```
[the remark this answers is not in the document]
demianw: Maybe we could call it ProbabilisticDecorator? Which decorates anything with a probability? Or just keep it as a probabilistic predicate but make it a three-parameter expression (functor, args, probability)

neurolang/frontend/tests/test_probabilistic_frontend.py
```diff
@@ -692,6 +695,87 @@ def test_query_based_pfact():
     assert_almost_equal(result, expected)
 
 
+def test_query_based_pfact_region_volume():
+    from ...expression_pattern_matching import add_match
+    from ...expression_walker import PatternWalker
+    from ...expressions import Constant, FunctionApplication, Symbol
+    from ...frontend.probabilistic_frontend import RegionFrontendCPLogicSolver
+    from ...logic import Conjunction, Implication
```
tgy, line 703: I put these imports here because, since we are in the probabilistic frontend tests, I should not have to import these IR classes. As the resolution of variable equalities of the form `x = f(g(x))` does not work, I have to use the trick below to convert `x = volume(y) / volume(z)` to `x = _f1_ / _f2_, _f1_ = volume(y), _f2_ = volume(z)`.

neurolang/frontend/tests/test_probabilistic_frontend.py
```diff
@@ -668,6 +671,111 @@ def test_solve_marg_query():
...
+            formula.functor == EQ
+            and isinstance(formula.args[1], FunctionApplication)
+            and formula.args[1].functor == Constant(operator.truediv)
+            and all(
+                isinstance(arg, FunctionApplication)
+                for arg in formula.args[1].args
+            )
+        )
+
+    class VolumeDivisionEqualityRewritter(PatternWalker):
```
demianw, line 716: this is specific to formula compositions. My question is why isn't this working directly. It should be well-translated and compiled by the RA translator. Where is the issue? This is a very specific fix which will not generalise and bring problems down the road.

neurolang/probabilistic/cplogic/program.py
```diff
@@ -6,8 +6,8 @@
 from ...exceptions import ForbiddenDisjunctionError, ForbiddenExpressionError
 from ...expression_pattern_matching import add_match
 from ...expression_walker import ExpressionWalker, PatternWalker
-from ...expressions import Constant, FunctionApplication, Symbol
-from ...logic import Implication, Union
+from ...expressions import MATMUL, Constant, FunctionApplication, Symbol
```
demianw, line 9: check my comment about me disagreeing with the way you handle the MATMUL constant

neurolang/probabilistic/cplogic/program.py
```diff
@@ -84,6 +91,34 @@ def within_language_prob_query(self, implication):
...
+        P(x) : y :- Q(x)
+
+    This is useful when the rule was defined at the frontend level using the
+    sugar syntax `(P @ y)[x] = Q[x]`.
+
+    """
+
+    @add_match(
+        Implication(
+            FunctionApplication(MATMUL(Symbol, ...), ...),
```
demianw, line 111: This seems to be sugar which is more appropriate to the frontend. Even if it works at the IR level it should be a model in the frontend. Check my PR https://github.com/NeuroLang/NeuroLang/pull/535 as an example of the architectural choice of sugar.

neurolang/probabilistic/query_resolution.py
```diff
@@ -1,16 +1,63 @@
...
+        det_antecedent = conjunct_formulas(impl.antecedent, eq_formula)
+        det_consequent = det_pred_symb(prob_symb, *impl.consequent.body.args)
+        det_rule = Implication(det_consequent, det_antecedent)
+        prob_consequent = ProbabilisticPredicate(
+            prob_symb, impl.consequent.body
+        )
+        prob_rule = Implication(prob_consequent, det_consequent)
+        self.walk(det_rule)
+        self.walk(prob_rule)
+        return impl
```
demianw, line 60: if you have changed impl then you should re-walk it, if not this is not a composable mixin

demianw (changes requested):
Many comments. Ignoring everything related to probabilistic facts as that's in PR #549 I assume. There are still reformatting / renaming diffs in this PR. This still forces me to deal with non-essential information in the PR, please be more careful in the future @tgy . Keep the renaming / styling out of the PR until all logic has been reviewed and approved.

neurolang/probabilistic/cplogic/program.py
tgy: Then maybe `TranslateProbabilisticQueryMixin`, which handles the `//` syntax for MARG queries should also be in the frontend module? You put it in the `neurolang/probabilistic/cplogic/program.py` module as well.

neurolang/expressions.py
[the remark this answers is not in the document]
tgy: fixed

neurolang/probabilistic/cplogic/program.py
demianw: You are right!!! @tgy let's merge it like this and then refactor both of them out, we should put "sugars" in the frontend. It was my bad before.

neurolang/probabilistic/expressions.py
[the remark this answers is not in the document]
tgy: agreed to leave that for later, opened https://github.com/NeuroLang/NeuroLang/issues/559 for posterity

neurolang/probabilistic/query_resolution.py
tgy: This has now been fixed. Note that I had to add a pattern for a Union containing a query-based probabilistic fact because the current pattern matching system makes it not convenient to return a Union of multiple formulas when walking a single formula. This leads to rules being walked multiple times (as the Union is re-walked when modified, by the `ExpressionWalker` mixin).

neurolang/frontend/tests/test_probabilistic_frontend.py
tgy: Has been removed now that this is working

neurolang/frontend/tests/test_probabilistic_frontend.py
tgy: has been removed

neurolang/probabilistic/cplogic/program.py
tgy: has been solved

tgy:
@demianw any idea why the test suite is not being ran on this PR?

demianw:
Nope, maybe you'll find insight here? https://app.circleci.com/pipelines/github/NeuroLang

neurolang/probabilistic/cplogic/program.py
```diff
@@ -84,6 +91,34 @@ def within_language_prob_query(self, implication):
         return self.walk(Implication(consequent, implication.antecedent))
 
 
+class TranslateQueryBasedProbabilisticFactMixin(PatternWalker):
```
demianw, line 94: I think that now you can move this to https://github.com/NeuroLang/NeuroLang/blob/master/neurolang/frontend/datalog/intermediate_sugar.py

neurolang/probabilistic/query_resolution.py
```diff
@@ -1,16 +1,86 @@
...
+
+    """
+
+    @add_match(
+        Union,
+        lambda union: any(
+            is_query_based_probfact(formula) for formula in union.formulas
+        ),
+    )
+    def union_with_query_based_pfact(self, union):
```
demianw, line 45: Do you need this if you have the Implication case?

demianw (changes requested):
Good job! Very minor changes

neurolang/probabilistic/query_resolution.py
tgy: I tried not to have this pattern. It does not work because the implication pattern returns TWO expressions. As we've seen before, our pattern matching system is not well suited to handle a case where we walk a Union of rules, one of which becomes several rules. We first walk the Union returned when walking the rule, and then the wrapper Union is modified and we re-walk the rules again. This results in the rules being walked multiple times. Attempting to do this, in this case, breaks the Forbidden Disjunction assumption of CP Logic program. If you have a better solution, I'm all ears

neurolang/probabilistic/cplogic/program.py
tgy: done

neurolang/probabilistic/query_resolution.py
demianw: OK It was a question. Good job

tgy: sure! i think we should come up with a strategy to handle these cases where one rule becomes multiple rules. i think it's the 3rd time that i have to work around this limitation with a pattern such as this one. i opened an issue for it a while ago, so no need to open another

demianw (approved):
It's great. Can you just add a test case where the stratification fails such that we know that we throw the good error? Basically there is a probabilistic predicate on a query-based probabilistic fact?

neurolang/probabilistic/query_resolution.py
demianw: You can check how @Jbat1Jumper solved it on `Fol2DatalogMixin` I think he solved well there.

demianw:
@tgy did you see my comment?

tgy:
good catch, this was not correctly handled, fixed

PR #550, opened 2020-11-09
Fix docstring for sphinx gallery for two-term relaxed CBMA conjunctive query

tgy:


tgy:
sphinx gallery compiled locally seems to work

demianw:
Can you provide a title without acronyms?

A description of the goals of the example such that people can understand what you are doing would be great

examples/plot_neurosynth_relaxed_tfidf.py
```diff
@@ -0,0 +1,150 @@
...
+    return ExplicitVBROverlay(
+        voxels, mni_t1_4mm.affine, p, image_dim=mni_t1_4mm.shape
+    )
+
+
+###############################################################################
+# Loading the database
+
+ns_database_fn, ns_features_fn = nilearn.datasets.utils._fetch_files(
+    "neurolang/frontend/neurosynth_data",
```
demianw, line 67: @tgy I think it's better to synch this to the other examples. This avoids downloading the dataset several times. Please change it to be in aggrement here.

examples/plot_neurosynth_relaxed_tfidf.py
```diff
@@ -0,0 +1,150 @@
...
+ns_database = set(
+    ns_database[["i", "j", "k", "id"]].itertuples(name=None, index=False)
+)
+
+ns_features = pd.read_csv(ns_features_fn, sep="\t")
+ns_docs = ns_features[["pmid"]].drop_duplicates().values
+ns_tfidf = pd.melt(
+    ns_features, var_name="term", id_vars="pmid", value_name="TfIdf"
+)
+ns_tfidf = ns_tfidf.loc[ns_tfidf.term.isin(["memory", "auditory"])]
```
demianw, line 103: @tgy It's kind of not great if we need to pre-filter the data. Is this because of speed?

tgy: Yes, I've added this for speed. Reason for that is that the expression for computing the relaxed tfidf probabilities is applied by pandas one row at a time. If you do the cross-product of all studies and terms, you get 46389588 rows. Computing the probabilities for all those rows is very long (haven't waited for the execution to finish because it was taking too long). Ideally we would compute probabilistic facts lazily (when solving query-based probfacts), only for those that are needed to solve the query. However, this has not been implemented yet. Another solution would be to make pandas solve this better, as it's an arithmetic formula on the columns, it should be able to compile that better? Not sure how to do it though, as I have not investigated.

demianw: The way to compute it "lazily" would be to do implement constant propagation, a simplified version of magic for non-recursive cases. What happens if you put directly the expression after the `@`? Except for the exponential it should go through the `StringArithmeticWalker` except of the exponential which is in pandas as `exp`. So extending the `StringArithmeticWalker` with a pattern matching case of `np.exp` to be converted to the string `exp`, should work.

```
import numpy as np
import pandas as pd
nn = np.random.randn(int(46e6), 1)
df = pd.DataFrame(nn, columns=['A'])
r = df.eval('1 / (1 + exp(A))')
```
the last operation takes 200ms in my computer.

examples/plot_neurosynth_relaxed_tfidf.py
tgy: fixed

neurolang/datalog/tests/test_cq_disj_to_named_ra.py
```diff
@@ -788,3 +790,91 @@ def test_equality_nested_builtin_application():
...
+        ),
+    )
+    assert result == expected
+
+
+def test_sigmoid():
+    x = Symbol("x")
+    z = Symbol("z")
+    P = Symbol("P")
+    EXP = Constant[Callable[[float], float]](
```
tgy, line 856: if the type of the callable is not fully specified, the translation does not work because the application of `EXP` is not treated as a builtin function application by the translator.

neurolang/frontend/query_resolution.py
```diff
@@ -366,7 +366,9 @@ def functions(self) -> List[str]:
         ]
 
     def add_symbol(
-        self, value: Union[fe.Expression, ir.Constant, Any], name: str = None
+        self, value: Union[fe.Expression, ir.Constant, Any],
+        name: str = None,
+        type_: Type = Unknown,
```
tgy, line 371: had to add this extra `type_` parameter to make it possible to specify the type of the value directly.

otherwise, had no way of enforcing that the `exp` symbol was typed as `Callable[[float], float]` and the translation to named RA was not working

neurolang/frontend/query_resolution.py
```diff
@@ -406,8 +408,12 @@ def add_symbol(
             value = value.expression
         elif isinstance(value, ir.Constant):
             pass
-        else:
+        elif type_ is Unknown:
             value = ir.Constant(value)
+        else:
+            value = ir.Constant[type_](
+                value, auto_infer_type=False, verify_type=False
```
tgy, line 415: not sure about `verify_type=False` here

neurolang/tests/test_expressions.py
```diff
@@ -316,3 +317,8 @@ class TestSymbol(expressions.Symbol):
 
     assert isinstance(TestSymbol.fresh(), TestSymbol)
     assert isinstance(TestSymbol[int].fresh(), TestSymbol[int])
+
+
+def test_numpy_ufunc_no_qualname_repr():
```
tgy, line 322: changes from other PR

neurolang/relational_algebra.py
```diff
@@ -434,6 +436,14 @@ def arithmetic_operation(self, fa):
             verify_type=False,
         )
 
+    @ew.add_match(FunctionApplication(Constant(numpy.exp), (...,)))
```
tgy, line 439: not sure `numpy.exp` is the right choice here. maybe `math.exp` would be more appropriate?

neurolang/frontend/query_resolution.py
demianw: Verify type here should be true

neurolang/frontend/query_resolution.py
demianw: I think that we should remove the `ir.Constant` from the type annotation. `ir.Constants` shouldn't even exist on the external API of the frontend.

demianw (changes requested):
Just two minor changes. Then if the tests pass (they aren't now) and the doc builds. We merge

neurolang/relational_algebra.py
demianw: maybe support both? @tgy

demianw (changes requested):
Good job!
Minor changes and please address the codacy issues

examples/plot_neurosynth_relaxed_tfidf.py
```diff
@@ -0,0 +1,147 @@
...
+with relaxed TFIDF thresholding on the Neurosynth database
+============================================================================
+
+This example first shows how to encode the Neurosynth Coordinate-Based
+Meta-Analysis (CBMA) database in NeuroLang, with probabilistic term-to-study
+associations using a sigmoid-based soft thresholding of TFIDF features.
+
+It then shows how a two-term conjunctive query can be expressed and solved to
+obtain an uncorreted forward inference map for studies associated with both
+terms 'auditory' and 'memory'.
```
demianw, line 13: Is this still correct? You are not doing 'auditory' and 'memory' anymore.

examples/plot_neurosynth_relaxed_tfidf.py
```diff
@@ -0,0 +1,147 @@
...
+SelectedStudy = nl.add_uniform_probabilistic_choice_over_set(
+    ns_docs, name="SelectedStudy"
+)
+
+###############################################################################
+# Probabilistic program and querying
+#
+# Compute a forward inference map for studies associated to both terms 'angular
+# gyrus' and 'semantic'. As per [1]_, "semantic processing is the most
+# consistent function that activates the angular gyrus".
```
demianw, line 114: good job finding an example that makes sense! However the result does not portray your hypothesis unless you run analyses to compare with all other functions in the paper. So a citation and an example like this is showing that we don't understand what we are querying.

demianw (changes requested):
The example still needs improving to make sense

demianw:
@tgy ping on this example. We should finish it soon-ish

PR #558, opened 2020-11-12
Probabilistic Choice Self-Join resolution through unification

tgy:
The current version of the code does not handle probabilistic choice self-joins
correctly. What currently happens is that a rule

    ans(x) :- P(x), P(y)

where `P` is a probabilistic choice predicate symbol, is translated by the
lifted optimisation function to the rule

    ans(x) :- P(x), P(y), y = x

Then, when this rule's hierarchicality is verified, it is not hierarchical.

This PR uses unification in the lifted optimisation to apply the variable
equalities and instead return the rule

    ans(x) :- P(x)

In the case where the unification completely removes one variable from the
antecedent that is not a free variable but a head variable, the equality is
maintained in the conjunctive antecedent to make sure the rule stays valid. If
the rule's consequent contains a variable that does not occur in the
antecedent, it's an invalid rule.

For example, the rule (where `P` is a probabilistic choice predicate)

    ans(x, y) :- P(x), P(y)

becomes

    ans(x, y) :- P(x), P(y), y = x

and then is unified as

    ans(x, y) :- P(x)

which is invalid. But, keeping the variable equality maintains the validity of
the rule.

### DIFF
--- a/neurolang/probabilistic/expression_processing.py
+++ b/neurolang/probabilistic/expression_processing.py
@@ -350,6 +350,11 @@ def lift_optimization_for_choice_predicates(query, program):
             unifier = UnifyVariableEqualities()
             rule = Implication(Symbol.fresh()(tuple()), query)
             query = unifier.walk(rule).antecedent
+            # keep head variable equalities to prevent the rule from becoming
+            # invalid due to a head variable not occurring in the antecedent
+            # by keeping these variable equalities, we make sure that the
+            # logical semantic of the rule is maintained (without modifying the
+            # query's head)
             kept_equalities = Conjunction(
                 (
                     eq
--- a/neurolang/probabilistic/expression_processing.py
+++ b/neurolang/probabilistic/expression_processing.py
@@ -350,11 +350,6 @@ def lift_optimization_for_choice_predicates(query, program):
             unifier = UnifyVariableEqualities()
             rule = Implication(Symbol.fresh()(tuple()), query)
             query = unifier.walk(rule).antecedent
-            # keep head variable equalities to prevent the rule from becoming
-            # invalid due to a head variable not occurring in the antecedent
-            # by keeping these variable equalities, we make sure that the
-            # logical semantic of the rule is maintained (without modifying the
-            # query's head)
             kept_equalities = Conjunction(
                 (
                     eq
--- a/neurolang/probabilistic/expression_processing.py
+++ b/neurolang/probabilistic/expression_processing.py
@@ -247,23 +247,23 @@ def construct_within_language_succ_result(provset, rule):
     return Constant[AbstractSet](provset.value.projection(*proj_cols))
 
 
-def group_preds_by_pred_symb(predicates, filter_set=None):
+def group_preds_by_functor(predicates, filter_set=None):
     """
-    Group predicates by their predicate symbol.
+    Group predicates by their functor.
 
-    An optional filter set of predicate symbols can be passed to only return
-    the ones in the set.
+    An optional filter set of functors can be passed to only return the ones in
+    the set.
 
     Parameters
     ----------
     predicates : iterable of predicates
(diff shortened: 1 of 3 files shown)


neurolang/probabilistic/expression_processing.py
```diff
@@ -345,6 +347,22 @@ def lift_optimization_for_choice_predicates(query, program):
...
         if len(added_equalities) > 0:
             query = Conjunction(query.formulas + tuple(added_equalities))
+            unifier = UnifyVariableEqualities()
+            rule = Implication(Symbol.fresh()(tuple()), query)
+            query = unifier.walk(rule).antecedent
+            # keep head variable equalities to prevent the rule from becoming
+            # invalid due to a head variable not occurring in the antecedent
+            # by keeping these variable equalities, we make sure that the
+            # logical semantic of the rule is maintained (without modifying the
+            # query's head)
```
demianw, line 357: I generally discourage this type of commenting in the middle of the code. Code should be declarative and readable.  If you want to document code sections like this, prioritise refactoring into a function and documenting the function's behaviour.

neurolang/probabilistic/tests/test_probabilistic_solvers.py
```diff
@@ -726,6 +727,27 @@ def test_empty_result_program(solver):
     assert testing.eq_prov_relations(result, expected)
 
 
+def test_program_with_probchoice_selfjoin(solver):
```
demianw, line 730: what happens if the probabilistic choice has more than one argument? Your algorithm is only targeted to this case, what if it is `P(x,y), P(z,w)`?

demianw (changes requested):
Strategy is good, I wonder if going first through equalities is not an unnecessary extra step and too custom-targeted at one-argument predicates. Wouldn't it just work to unify all choice predicates with the same predicate symbol directly?

@tgy: update by unifying I mean given
```
ans(x, y) :- P(x), P(y)
```
take all the variables across all appearances of the predicate and produce the replacement 
`{x:y}` for instance. If there is a constant, then variable should always be replaced by constant. You can merge or compose the substitutions across all appearances of the predicate with `merge_substitutions` or `compose_substitutions`. If there is no substitution possible then the antecedent should be `false`

neurolang/probabilistic/tests/test_probabilistic_solvers.py
tgy: the algorithm was designed to work for an arbitrary number of variables. i've added a test for the `P(x, y), P(z, w)` case

neurolang/probabilistic/expression_processing.py
tgy: removed

tgy:
I've updated the algorithm to try to integrate your proposal

neurolang/probabilistic/tests/test_probabilistic_solvers.py
```diff
@@ -726,6 +728,50 @@ def test_empty_result_program(solver):
...
+    cpl = CPLogicProgram()
+    cpl.add_probabilistic_choice_from_tuples(
+        P,
+        [(0.2, "a", "b"), (0.8, "b", "c")],
+    )
+    cpl.add_probabilistic_facts_from_tuples(
+        Q,
+        [(0.6, "a"), (0.8, "b")],
+    )
+    rule = Implication(R(x, y, z, w), Conjunction((Q(x), P(x, y), P(z, w))))
```
demianw, line 765: looks great! Does a case like `P(x, y), P(z, x)` work? If it works and it's tested, then it's a merge

demianw (changes requested):
Small question

neurolang/probabilistic/tests/test_probabilistic_solvers.py
tgy: test added, works

PR #562, opened 2020-11-12
[fix] maintain my_symbol.is_fresh in ReplaceExpressionWalker

tgy:


tgy:
stumbled upon this in my tests and it was causing an issue

@demianw i think you initially wanted to do this because the `changed` variable was defined but not used (codacy should have caught that 👁️).

neurolang/expression_walker.py
```diff
@@ -270,6 +270,7 @@ def replace_expression(self, expression):
         for arg in args:
             if isinstance(arg, Expression):
                 new_arg = self.walk(arg)
+                changed |= new_arg is not arg
             elif isinstance(arg, tuple):
                 new_arg, _ = self.process_iterable_argument(arg)
```
demianw, line 275: This will not work if there is a change inside of the tuple will it?

demianw (changes requested):
Good catch! Little point you missed. Maybe add a test for it?

neurolang/tests/test_expression_walker.py
```diff
@@ -192,3 +192,19 @@ def test_convert_to_lambda():
...
     assert l3(x=3) == 2
     assert args == {'x'}
+
+
+def test_expression_walker_keep_fresh():
+    s = S_.fresh()
+    assert s.is_fresh
+    walker = expression_walker.ExpressionWalker()
+    result = walker.walk(s)
+    assert result.is_fresh
```
demianw, line 202: This should check that the symbol is exactly the same not only `is_fresh` as a fresh symbol can appear many times in an expression @tgy

neurolang/expression_walker.py
tgy: strangely, it works.. but will still use the `change` output of `process_iterable_argument`, good point

neurolang/tests/test_expression_walker.py
tgy: fixed

PR #563, opened 2020-11-13
[fix] flattening of self-join with constants

tgy:
Stumbled upon this in my CBMA two-term conjunctive query testing.

Note that this branch is based on
https://github.com/NeuroLang/NeuroLang/pull/562, which should be merged first.

Let the single-rule program be

    P(x) :- R(x)

and the query be `ans :- P(a), P(b)`.

The flattening of the conjunction `P(a), P(b)` currently gives

    R(a), x = a, R(b), x = b

which is not what we want.

That happens because the flattening does not handle well the self-join with
constants.

My strategy for fixing this is to create a new mixin that also freshens all
head variables of the rules of the program before the application of the MGU.

Now, the flattening of `ans :- P(a), P(b)` is

    R(_f1_), _f1_ = a, R(_f2_), _f2_ = b


demianw:
This seems to be a shattering problem not a flattening problem. Could you check that @tgy ?

demianw:
I don't understand why the equalities need to be added instead of just replacing the variables in the head by the terms used when the head appears in the body of another rule.

tgy:
You're right, I think the issue here was that the `x = a` and `x = b` equalities were added, which was not wanted.

tgy:
Now the flattening correctly outputs `R(a), R(b)`

PR #564, opened 2020-11-13
Add two-term CBMA conjunctive query test of probabilistic frontend

tgy:


tgy:
Note that this requires https://github.com/NeuroLang/NeuroLang/pull/563 to pass

demianw (approved):
Good job! We need to merge the previous PRs

PR #565, opened 2020-11-13
[fix] allow multiple equalities in query hierarchicality test

tgy:


neurolang/probabilistic/dichotomy_theorem_based_solver.py
```diff
@@ -100,6 +100,9 @@ def is_hierarchical_without_self_joins(query):
 def extract_atom_sets_and_detect_self_joins(query):
     has_self_joins = False
     predicates = extract_logic_atoms(query)
+    predicates = set(
+        pred for pred in predicates if isinstance(pred.functor, Symbol)
```
tgy, line 104: only consider predicates whose functor is a relational symbol

tgy:
the issue is that multiple equalities (or other builtin) in the query leads to the query to be considered as having self joins

neurolang/probabilistic/dichotomy_theorem_based_solver.py
```diff
@@ -100,6 +100,9 @@ def is_hierarchical_without_self_joins(query):
 def extract_atom_sets_and_detect_self_joins(query):
     has_self_joins = False
     predicates = extract_logic_atoms(query)
+    predicates = set(
+        pred for pred in predicates if not isinstance(pred.functor, Constant)
```
tgy, line 104: only consider predicates whose functor is not constant

demianw:
This is not a fix @tgy builtins where left out of probabilistic predicates explicitly because the probabilistic semantics of any given builtin are impossible to guarantee. You can enable equality.

tgy:
Good point. Only allowing equalities now. Also added a test for other builtins.

PR #566, opened 2020-11-13
Unification after lifted query optimisation of self-joined probabilistic choices

tgy:


demianw:
Can you please describe the formal algorithm that you implemented in the description @tgy ?  It's hard to understand from your code and it hampers the distinction between the formalism behind your algorithm and the algorithm that might have a bug or a difference with the semantics.

tgy:
When given a (flattened) query of the form
```
ans(v) :- A(v, s1), T(t1, s1), T(t2, s2), S(s1), S(s2)
```
where `S` is a probabilistic choice relation, the current version of the algorithm first establishes that `s1 = s2` through the "lifted query optimisation of probabilistic choices", which gives
```
ans(v) :- A(v, s1), T(t1, s1), T(t2, s2), S(s1), s1 = s2
```
**This gives the right probability estimations.** Which means that the current version of the algorithm is working.

However, when translated to Named Relational Algebra, this expression will construct a _very large_ relation when processing the natural joins for `A(v, s1), T(t1, s1), T(t2, s2)`. Note that, in this particular example, there is a _cross product_ that is calculated for `T(t1, s1), T(t2, s2)`.

What this proposes is adding a _unification_ step that applies the substitution `{s1 / s2}` on the query, giving the CQ
```
ans(v) :- A(v, s1), T(t1, s1), T(t2, s1), S(s1)
```
thereby avoiding the cross product mentioned above.

An important point is that "extended projection" probabilistic choice equalities are maintained in the body of the query whenever they pertain to a head variable of the query. For example, if the query is
```
ans(x, y) :- A(x, y), x = y
```
which is unified to
```
ans(x, x) :- A(x, x)
```
Instead, only the body is unified and the `x = y` equality is maintained in the body:
```
ans(x, y) :- A(x, x), x = y
```
which leads to the same resolution, but also projects the new column `y` with the `x` value of the resulting set.

Maintaining the query's head makes the processing easier in the shattering later.

This is my proposed strategy. If you have a better proposal, I'm all ears.

demianw:
If this is your proposal, which is OK. There is no shattering involved, the title is misleading and hence without the description, the code is very hard to follow.

tgy:
The title did not meant to say that shattering was involved. It was instead saying that this was adding a "unification" step between the "flattening" and "shattering" steps. However, I realise from your confusion that this could have been confusing. I have modified the title. Let me know if this makes it more explicit.

demianw:
It's good. The new title plus the description of the take make it easier for me to review the logic of the code. Thanks

neurolang/datalog/expression_processing.py
```diff
@@ -868,6 +868,33 @@ def _get_between_symbs_equality_substitutions(eq_set):
...
+    query = enforce_conjunctive_antecedent(query)
+    extractor = UnifyVariableEqualitiesMixin._Extractor()
+    extractor.walk(query.antecedent)
+    replacer = ReplaceSymbolWalker(extractor.substitutions)
+    conjuncts = set(
+        replacer.walk(formula)
+        for formula in query.antecedent.formulas
+        if not is_equality_between_symbol_and_symbol_or_constant(formula)
+    )
+    if keep_head_var_eqs:
```
demianw, line 881: Datalog does not have named semantics. This seems unnecessary in a datalog-pure case. If we need to keep these semantics maybe it's the wrong module

neurolang/logic/compare.py
```diff
@@ -0,0 +1,56 @@
+from ..expression_pattern_matching import add_match
+from ..expression_walker import ExpressionWalker
+from ..expressions import Definition, Expression
+from . import NaryLogicOperator
+
+
+class LogicalComparison(Definition):
```
demianw, line 7: Logical comparison is an extremely hard problem. Hence this whole module, through naming, misleads anyone using it further than the road. This seems to be only for testing purposes. I think that Nikita made comparisons possible through an MGU strategy on one of his modules you could check that.

Also you use this for testing only which, according to me, complexifies a lot understanding what's happening when a test fails. Try to design tests where verifying the validity of the test doesn't require a new module with complex functionality

demianw (changes requested):
Careful, the way you enforce conjunctions instead of trusting the whole system of preprocessing through mixins and expressions that we have makes the code you produce very hard to extend, for instance in terms of adding negation.

neurolang/datalog/expression_processing.py
```diff
@@ -868,6 +868,33 @@ def _get_between_symbs_equality_substitutions(eq_set):
         return {symb: chosen_symb for symb in iterator}
 
 
+def unify_query_vareqs(query, keep_head_var_eqs=False):
+    query = enforce_conjunctive_antecedent(query)
+    extractor = UnifyVariableEqualitiesMixin._Extractor()
```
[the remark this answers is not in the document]
tgy: Fixed

neurolang/datalog/expression_processing.py
```diff
@@ -868,6 +868,33 @@ def _get_between_symbs_equality_substitutions(eq_set):
         return {symb: chosen_symb for symb in iterator}
 
 
+def unify_query_vareqs(query, keep_head_var_eqs=False):
+    query = enforce_conjunctive_antecedent(query)
```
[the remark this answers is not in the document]
tgy: This is not a *validation*. This function transforms the rule's antecedent to make it a `Conjunction` even though it is a single predicate antecedent. For example, the rule `Q(x) :- P(x)` is transformed to `Q(x) :- Conjunction((P(x),))`. However, the rule `Q(x) :- P(x), Z(x)` is left untouched, as the antecedent is already a conjunction. Maybe I'm missing something but I'm not sure why this would be an issue for rules with negation or aggregation. In the case of negation, wrapping a negative literal with a conjunction does not change the semantic of the rule. In the case of aggregations, I don't see the issue with this modification as the consequent of the rule is left untouched, and similarly wrapping the single predicate antecedent of an aggregation rule does not change the semantic of the rule. Maybe I am not understanding what you are trying to convey. Could you please elaborate?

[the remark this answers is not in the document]
demianw: Nope you got it right. I got confused by the use of the word `enforce`. It does work.

neurolang/datalog/expression_processing.py
```diff
@@ -868,6 +868,33 @@ def _get_between_symbs_equality_substitutions(eq_set):
...
+    extractor = UnifyVariableEqualitiesMixin._Extractor()
+    extractor.walk(query.antecedent)
+    replacer = ReplaceSymbolWalker(extractor.substitutions)
+    conjuncts = set(
+        replacer.walk(formula)
+        for formula in query.antecedent.formulas
+        if not is_equality_between_symbol_and_symbol_or_constant(formula)
+    )
```
demianw, line 887: What happens if you have an antecedent `Q(x, y), z=y, y=1`, is there a possibility that this will end up as `Q(x,z), y=1`?

neurolang/datalog/expression_processing.py
```diff
@@ -868,6 +868,33 @@ def _get_between_symbs_equality_substitutions(eq_set):
...
+    if keep_head_var_eqs:
+        head_symbols = set(
+            arg for arg in query.consequent.args if isinstance(arg, Symbol)
+        )
```
demianw, line 891: will this work with an aggregation case?

neurolang/logic/compare.py
tgy: Ok, I did not mean to make a logical _equivalence_ verification, but simply a comparator of logic expressions that took into account the commutativity n-ary logic operators. I believe this simple functionality is not as complex as you make it sound, and that it could be benificial in making the tests more readable. I will give you an example.

This test, to me, is hard to write and hard to read. Writing all those assertions make the tests longer and require a larger cognitive load for the developer who writes / reads the test.
```
    expected = Union(
        (
            Implication(
                R(x, a),
                Conjunction((P(a), Q(a, x), Q(a, a), Z(b), Z(a))),
            ),
        )
    )
    result = program.intensional_database()[R]
    assert isinstance(result, Union)
    assert len(result.formulas) == 1
    assert isinstance(result.formulas[0], Implication)
    assert isinstance(result.formulas[0].antecedent, Conjunction)
    assert result.formulas[0].consequent == expected.formulas[0].consequent
    assert set(result.formulas[0].antecedent.formulas) == set(
        expected.formulas[0].antecedent.formulas
    )
```
while
```
    expected = Union(
        (
            Implication(
                R(x, a),
                Conjunction((P(a), Q(a, x), Q(a, a), Z(b), Z(a))),
            ),
        )
    )
    result = program.intensional_database()[R]
    assert logic_exps_equal(result, expected)
```
is much shorter and easier to read.

I believe this _simple_ logical expression comparator could be useful for our tests. Maybe not in the `logic` module and more into a `test_utils` module.

But of course, you decide. If your final decision is to remove it, I will remove it.

neurolang/datalog/expression_processing.py
tgy: I've added tests to verify this. I can confirm that `ans(x, y) :- Q(x, y), z = y, y = 1` is correctly translated to `ans(x, 1) :- Q(x, 1)` and that `ans(x, y, z) :- Q(x, y), z = y, y = 1` is correctly translated to `ans(x, 1, 1) :- Q(x, 1)`.

neurolang/logic/compare.py
demianw: The main issue is that you are using words with a precise technical meaning and definition and not respecting their semantics.  Two first order logic expressions are `equivalent` (or equal but that is seldom used) when the set issued from their logical inference is the same. Hence, your naming is misleading. In general you need to be more careful on how you use technical words (such as the previous title of this PR).

I agree with the `test_utils` module idea, it's a good one. If you manage to make the name more precise, and document it, then it's OK.

neurolang/datalog/expression_processing.py
demianw: The problem for me is that as you have not formally defined your algorithm, and the code is spread across many functions it's very hard to understand that it will work in all cases. Could you please formalise the algorithm? If not we are in danger of adding bugs to the solver.

demianw: @tgy this is not resolved, you algorithm it's *described* above, it's not a formalised pseudocode which allows to analyse and hopefully prove that the algorithm works.

neurolang/logic/compare.py
tgy: I moved the module to `neurolang.utils.testing.logic_commutative_comparison`. Let me know if that works for you.

demianw: Good idea! Maybe just `neurolang.utils.testing.logic` as there might be other logical utilities that might want to be added there, add the `__all__` variable to only export the function and not the class. Also add documentation to the function describing the precise check that you are doing on the logic expression it's still unclear.

neurolang/datalog/expression_processing.py
tgy: I'm not sure what to do in this case
```python
def test_aggregation():
    rule = Implication(
        P(x, Constant(sum)(y, z)),
        Conjunction(
            (
                R(x, y, z),
                EQ(z, a),
            )
        ),
    )
    program = DatalogWithVariableEqualityUnification()
    program.walk(rule)
    expected = Union(
        (
            Implication(
                P(x, Constant(sum)(y, a), Q(x, a)),
                R(x, y, a),
            )
        )
    )
    result = program.intensional_database()[P]
    assert logic_exp_commutative_equal(result, expected)
```
I checked in the aggregation implementation and could not find examples of aggregation + equality

neurolang/datalog/expression_processing.py
tgy: **Attempt to formalise the algorithm**

**Goal of the algorithm**: translate a rule based on variable equalities
present in its antecedent.
For example, a rule `ans(x, y) :- x = 2, y = 3, R(x)` should be translated to
`ans(2, 3) :- R(2)`.

**Strategy of the algorithm**: first extract the variable equalities using the
`ExtractVariableEqualities` walker, then apply the extracted variable
equalities to the rule's consequent and antecedent using a
`ReplaceSymbolWalker`.

**Notes**:

- The `ExtractVariableEqualities` walker has to handle all possible
  combinations of equalities, such as `y = 2 & y = 3` which will result in
  the antecedent becoming `Constant(False)` or such as `y = 1, y = z` which
  will result in two symbol replacements: `{y / 1, z / 1}`.
- The `unify_var_eqs` function has the option of leaving the consequent of the
  rule untouched (`keep_head_symb_vareqs`), meaning that the variables in the
  head of the rule are not going to be replaced. By default, everything is
  replaced. This option can be useful for when translating a rule to named
  relational algebra. By keeping the antecedent equalities that create new
  columns for head variables, the rule can simply be passed to Named Relational
  Algebra translator and it will result in the desired set. Another approach
  for that would be to translate the whole rule and then compare the old (with
  all head variables) and new (after translation) consequent's arguments to
  re-add the equalities. I think it's simpler to just do it here, by providing
  this option.

**Pseudo-code**

Here, I'm assuming `ExtractVariableEqualities` does its job well and takes care
of all cases of combinations of variable equalities (as mentioned in the
_Notes_ section above).

```
function unify_var_eqs
arguments
    rule
        a datalog-like implication rule with variable equalities in its
        antecedent
    keep_head_symb_vareqs
        a boolean argument (default to False) that enables the option of
        keeping the consequent (head) of the rule untouched
begin
    # mapping from variable to constant or another variable
    var_eqs := extract_variable_equalities(rule)
    # get formulas without variable equalities because variable equalities
    # have been used for substituting variables and are not needed anymore
    # as the semantic of the rule without equalities now emcompasses the
    # semantic of the variable equalities
    antecedent := all formulas in the rule that are not vareqs
    antecedent := substitute variables using var_eqs in antecedent
    if keep_head_symb_vareqs
        head_symbs := get_rule_head_symbs(rule)
        # get formulas x = something, where x is a head symbol
        head_symb_eqs := variable equality formulas defining head symbols
        antecedent := conjunct antecedent and head_symb_eqs
    else
        consequent := substitute variables using var_eqs in rule's consequent
    new_rule := create new rule based on new consequent and antecedent
    return new_rule
end
```

neurolang/datalog/expression_processing.py
demianw: You can try to come up with one, for instance
```
A[x, sum(y)] :- B(x), y = 1
```
should count the number of distinct elements in B(x)

neurolang/datalog/expression_processing.py
demianw: An issue in you algorithm is that if you add back the variable equalities with the head, you have to be sure to apply to those equalities the same substitutions as those to the antecedent. Same for the head variables. If not you might find yourself an equality with a variable that you have substituted out. So probably the substitutions should be done on the new_rule directly

demianw: I think that "keeping the head untouched" breeds for a problem where, in the body, you can substitute out the variables on the head.

neurolang/datalog/expression_processing.py
demianw: Your expected implication is syntactically incorrect @tgy

demianw: It's perfectly reasonable that your algorithm doesn't work or is not pertinent for aggregations. In this case, in which is only pertinent for the cases we are handling in the probabilistic semantics, which do not include aggregation, maybe requires you to move the code there. Or just throw an exception if the head contains something which is not a symbol as a parameter.
The main issue is that you write your code and your names as if they work for every case, this leads to a confusing API.

neurolang/logic/compare.py
tgy: Done

neurolang/datalog/expression_processing.py
tgy: Alright, I removed this functionality now and I'm dealing with the query's head variables in the probabilistic solver. I think this is cleaner, you are right.

neurolang/datalog/expression_processing.py
tgy: Raising an exception now when trying to unify variable equalities in a rule with aggregation.

neurolang/datalog/aggregation.py
```diff
@@ -26,41 +26,21 @@
...
-def is_aggregation_rule(rule):
-    return is_aggregation_predicate(rule.consequent)
-
-
-def is_aggregation_predicate(predicate):
-    return any(
-        isinstance(arg, AggregationApplication) for arg in predicate.args
-    )
-
-
```
tgy, line 63: I had to move these to `datalog.expressions` and `datalog.expression_processing` because an otherwise *circular* dependency

neurolang/datalog/expression_processing.py
tgy: Fixed

neurolang/datalog/expression_processing.py
```diff
@@ -745,7 +745,9 @@ def is_rule_with_builtin(rule, known_builtins=None):
 
 
 def remove_conjunction_duplicates(conjunction):
-    return Conjunction(tuple(set(conjunction.formulas)))
+    return maybe_deconjunct_single_pred(
+        Conjunction(tuple(set(conjunction.formulas)))
+    )
 
 
 def iter_disjunction_or_implication_rules(implication_or_disjunction):
```
demianw, line 753: isn't this a simplified version of `extract_logic_predicates` or `extract_logic_atoms`? If not what's the difference and why do we need a new function?

neurolang/datalog/expression_processing.py
```diff
@@ -868,29 +880,59 @@ def _get_between_symbs_equality_substitutions(eq_set):
...
-        extractor = UnifyVariableEqualitiesMixin._Extractor()
+        extractor = ExtractVariableEqualities()
         extractor.walk(implication.antecedent)
         replacer = ReplaceSymbolWalker(extractor.substitutions)
-        consequent = replacer.walk(implication.consequent)
-        conjuncts = tuple(
-            replacer.walk(formula)
-            for formula in implication.antecedent.formulas
-            if not is_equality_between_symbol_and_symbol_or_constant(formula)
+        antecedent = Conjunction(
```
demianw, line 906: somehow feels that this list by extension and the `ExtractVariableEqualities` could be summarised in a simple for loop or in one walker as they are precisely complementary, or am I missing something?

neurolang/utils/testing/logic.py
```diff
@@ -0,0 +1,67 @@
...
+        return all(
+            any(
+                self.walk(LogicCommutativeComparison(f1, f2))
+                for f2 in second.formulas
+            )
+            for f1 in first.formulas
+        )
+
+
+def logic_exp_commutative_equal(exp1, exp2):
```
demianw, line 62: can you please add documentation on what you are exactly checking here? The title is still not clear.

demianw (changes requested):
Good job, small changes

neurolang/datalog/expression_processing.py
tgy: It was for iterating on rules in an expression where the expression was either a rule or a disjunction of rules. But I can remove it because it's used in only one place now. Fixed

neurolang/datalog/expression_processing.py
tgy: I created two classes for modularity purposes. To be able to extract variable equalities without unifying them if need be. But I can merge all of this if you want. I was attempting to follow your principle of modularising things in separate walkers. You decide.

neurolang/utils/testing/logic.py
tgy: done

neurolang/datalog/expression_processing.py
tgy: also, keeping `ExtractVariableEqualities` in a walker makes it possible to extend it later if we want

demianw: The modularity idea is a good one. The issue for me is the unclear semantics. Which equalities from the extractor won't be in the `extractor.substitutions` argument and detected by your filter in line 902? If there are none, then the code is not reflecting this consistency.
Also, does the extractor know about substitutions? It shouldn't, substitutions have a directionality, equalities do not. This also makes for inconsistency between naming and behaviour. A better name would be `extracted_equalities`. Specifically, if you are going to modularise, then the module only knows what it nows and not it's ulterior purpose.  Or you might copy the idea of `ExtractLogicPredicates` and make a walker that walks returning a set.

Finally the implementation of `is_equality_between_symbol_and_symbol_or_constant` does not correspond with its name as a constant/constant equality will also be detected by this function, which also brings unclear semantics and possibly bugs. As a rule with an equality `1=2` is valid, should make the rule always false, and your processing will not respect this behaviour.

tgy: You're right. I decided to merge them into a single walker because there is no short term value in taking the time to implement this modularity.

tgy: also renamed `is_equality_between_symbol_and_symbol_or_constant` to `is_variable_equality`, enforcing the presence of a `Symbol` in one of the args

neurolang/datalog/expression_processing.py
```diff
@@ -926,6 +841,77 @@ def unsupported_unification_for_aggregation(self, rule_with_aggregation):
...
+                const = next(
+                    term for term in eq_set if isinstance(term, Constant)
+                )
+                update = {
+                    term: const for term in eq_set if isinstance(term, Symbol)
+                }
+            else:
+                iterator = iter(eq_set)
+                chosen_symb = next(iterator)
+                update = {symb: chosen_symb for symb in iterator}
```
demianw, line 858: this is great! I think if you sort symbols lexicographically (replaced < replacement), this will make the algorithm more stable and some times a bit more optimal (even if sometimes not)

neurolang/datalog/expression_processing.py
```diff
@@ -926,6 +841,77 @@ def unsupported_unification_for_aggregation(self, rule_with_aggregation):
...
+            and any(isinstance(arg, Symbol) for arg in predicate.args)
+        )
+        eq_sets = list()
+        for equality_predicate in equality_predicates:
+            if all(isinstance(arg, Symbol) for arg in equality_predicate.args):
+                UnifyVariableEqualities.add_equality_with_symbol(
+                    eq_sets, *equality_predicate.args
+                )
+            else:
+                symb = next(
```
demianw, line 877: Isn't this nicer as a sigle if statement and two options for making assignments? Using two filtered iterators combined with a next for a case which has only two arguments seems like overkill.

demianw (changes requested):
Two last changes and we are done!

neurolang/datalog/expression_processing.py
tgy: done by replacing all symbols by the maximum (lexicographically speaking) symbol

neurolang/datalog/expression_processing.py
tgy: done

PR #569, opened 2020-11-16
New provenance operation: Difference

gzanitti:


neurolang/relational_algebra.py
```diff
@@ -113,6 +113,14 @@ def __init__(self, relation_left, relation_right):
     def __repr__(self):
         return f"[{self.relation_left}" f"\N{JOIN}" f"{self.relation_right}]"
 
+class LeftNaturalJoin(RelationalAlgebraOperation):
+    def __init__(self, relation_left, relation_right):
+        self.relation_left = relation_left
+        self.relation_right = relation_right
+
+    def __repr__(self):
+        return f"[{self.relation_left}" f"\N{JOIN}" f"{self.relation_right}]"
```
demianw, line 122: I think the symbol for the left join is on unicode

demianw (changes requested):
Looks good! I think we are missing the cases where the set differences between left and right are non-empty

gzanitti:
> Looks good! I think we are missing the cases where the set differences between left and right are non-empty

@demianw But that's exactly the test case that we have. If you look at the second row of each set, they don't match, that's why the final probability of the second row has the values of `left`. Am I missing something?

PR #570, opened 2020-11-17
Improved tutorial

demianw:


doc/tutorial.rst
```diff
@@ -92,6 +95,45 @@ or in a less verbose manner:
...
+
+  with neurolang.scope as e:
+      e.odd[e.x] = ~e.even[e.x]
+
+      res = neurolang.query(
+          e.x, 
+          between_one_and_ten(e.x) & e.odd(e.x)
+      )
+
+we will obtain the set of tuples :code:`{(1,), (3,), (5,), (7,), (9,)}`
```
LouisRouillard, line 135: good example, as a matter of fact I think it could be clearer to make simple arithmetic examples like this before moving on to neuroscience-related ones, otherwise this is a bit too much to take at once

doc/tutorial.rst
```diff
@@ -111,6 +153,129 @@ which in neurolang is expressed as
    with neurolang.scope as e:
        e.max_population_per_country[e.country, e.max(e.pop)] = e.population_per_country_province(e.country, e.province, e.pop)
 
+
```
LouisRouillard, line 156: the example on population is a bit too verbose in its mathematical formulation, and not neuroscience-related, maybe this could be made more synthetic ?

doc/tutorial.rst
```diff
@@ -40,7 +40,10 @@ which we formalise in python as:
...
 
   with neurolang.scope as e:
-      e.left_hemisphere_gyrus[e.x] = e.destrieux_atlas(e.l, e.x) & e.startswith('L G', e.l)
+      e.left_hemisphere_gyrus[e.x] = (
+          e.destrieux_atlas(e.l, e.x) &
+          e.startswith('L G', e.l)
+      )
 
 
 the full example is in our gallery in :ref:`sphx_glr_auto_examples_plot_load_destrieux_left_hemisphere_gyri.py`.
```
LouisRouillard, line 49: dangling "Negation can also..." phrase !

doc/tutorial.rst
```diff
@@ -92,6 +95,45 @@ or in a less verbose manner:
       e.left_hemisphere_region[e.x] = e.left_hemisphere_sulcus(e.x) | e.left_hemisphere_gyrus(e.x)
```
LouisRouillard, line 96: this disjunction example is very clear

doc/tutorial.rst
```diff
@@ -111,6 +153,129 @@ which in neurolang is expressed as
...
+
+Is equivalent to 
+
+.. code-block:: python
+   
+    with neurolang.scope as e:
+        e.central[e.x] = Destrieux["Central Sulcus", e.x]
+
+for an example of application see :ref:`sphx_glr_auto_examples_plot_sulcal_queries.py` .
+
```
LouisRouillard, line 196: This example needs reworking, because in this very setup I'm not sure the syntactic sugar notation is actually easier to understand. Or at least the interest of it is not put forward enough

doc/tutorial.rst
```diff
@@ -111,6 +153,129 @@ which in neurolang is expressed as
...
+    with neurolang.scope as e:
+        e.most_anterior[e.r] = (
+            Destrieux[..., e.r] &
+            ~neurolang.exists(
+                e.q,
+                Destrieux(..., e.q) & (e.r != e.q) &
+                e.anterior_of(e.q, e.r)
+            )
+        )
+
```
LouisRouillard, line 220: actually it has already been pointed out that all predicates in the body were featuring only existential rules, I think this example would be more striking if we used "for all q" in stead of "non exists q". Otherwise the message is mixed here with negation.

doc/tutorial.rst
```diff
@@ -111,6 +153,129 @@ which in neurolang is expressed as
...
+        )
+  
+
+Probabilistic Semantics in Neurolang
+------------------------------------
+
+One of the main features of NeuroLang are its probabilistic semantics. Specifically, NeuroLang can represent discrete random variables and execute inference tasks.
+For a more detailed description of probabilistic logic languages please see the introduction of De Raedt et al [deraedt2015]_. Let's present an example:
+
+In neuroimaging meta-analysis, a main question is the probability that an article containg a specific term, such as "auditory" reports a voxel in the brain as active.
```
LouisRouillard, line 228: then again, I 'm not sure this should start straight on with a neuroscience-related example, but then again it is just a question of opinion.

doc/tutorial.rst
```diff
@@ -111,6 +153,129 @@ which in neurolang is expressed as
...
+    &\vdots\\
+    \mathbf{v_V}, \mathbf{s} &\sim \operatorname{Bernoulli}\left(\theta_{v_V, \mathbf{s}}\right)\\
+
+where :math:`\mathbf{s}` is a chosen study amongs those in the set :math:`S_1,\ldots,S_n`; :math:`\mathbf{t_i},\mathbf{s}` is the probability that the i-th term  is present in study *s*;
+and :math:`\mathbf{v_j}, \mathbf{s}` is the joint probability that the j-th voxel is reported active in study *s*. Our goal then is to compute the probability, that a voxel `j` is active
+if a certain term :math:`t_i` is reported in the study :math:`P(\mathbf{v_j}|\mathbf{t_i})`. In the following NeuroLang snippet we setup the computation of this conditional probability as well as
+of the marginal probability that a term has been mentioned in a study:
+
+.. code:: python
+
```
LouisRouillard, line 249: the loading of the TermInStudy, random choices and such should maybe be explicit *in the Python script* otherwise the reader is just wondering "wait where do my probabilities come from", "why is there a single study", etc...

doc/tutorial.rst
LouisRouillard: or at least not a meta-analysis example. Otherwise it would be cool to explicit a bit the goal of the field, because I'm not sure this is common knowledge in neuroscience ?

LouisRouillard (commented):
A good starting point, but I think it globally should be made more user-friendly, explaining some concepts in layman terms, and starting off softly using basic arithmetic examples before switching to moire interesting neuroscience-related examples.

doc/tutorial.rst
demianw: The problem with `for all q` is that unless we implement the implication, as an operator the syntax is very hard to read, I find:
```
forall q: ~Destrieux(_, q) | ((e.r != q) | e.anterior_of(e.q, e.r))
```

PR #578, opened 2020-11-20
Negation in XRewrite

gzanitti:
Fix error in XRewrite when trying to process negations.

### DIFF
--- a/neurolang/datalog/ontologies_rewriter.py
+++ b/neurolang/datalog/ontologies_rewriter.py
@@ -4,7 +4,7 @@ from ..expression_walker import (
     add_match,
 )
 from ..expressions import Symbol
-from ..logic import Constant, Implication, NaryLogicOperator
+from ..logic import Constant, Implication, NaryLogicOperator, Negation
 from ..logic.expression_processing import ExtractFreeVariablesWalker
 from ..logic.transformations import CollapseConjunctions
 from ..logic.unification import apply_substitution, most_general_unifier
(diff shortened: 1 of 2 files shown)


gzanitti:
Reminder:
- Compromise solution that allows to solve negations by means of XRewriter avoiding to evaluate them (at the moment it fulfills the requirements of our examples).
- We should evaluate the case where the unification is done between two negations.

I will convert the PR into a draft until then.

demianw:
@gzanitti can we add a test for the negation of an existential rule and the corresponding exception?

gzanitti:
> @gzanitti can we add a test for the negation of an existential rule and the corresponding exception?

The test added in this PR, fulfills the necessary condition, it only remains to define how to treat it. If we throw an exception every time there is a negated existential, all the examples we are already working on will stop running. Even if the negation does not influence the result.

I believe that I should throw the exception only in the case where, given the removal of the negation in the atom, the term has a constraint to unify with. What do you think?

demianw:
Can you give me a little code example?

gzanitti:
@demianw the tests fail because NeuroSynth is not installed. Any ideas?

`ImportError: Neurosynth not installed in the system`

demianw:
@gzanitti can we have the status of this? We need to merge this or it has been solved elsewhere?

gzanitti:
> @gzanitti can we have the status of this? We need to merge this or it has been solved elsewhere?

Should be solved in another PR, but let me confirm.

PR #579, opened 2020-11-23
Right operators in the frontend now work

demianw:


demianw:
@tgy this might help with the building of your operation now

PR #580, opened 2020-11-24
Check for __qualname__ presence in Constant.__repr__

tgy:
Stumbled upon this code assumption of the `__qualname__` attribute being
present for all `Constant` values. Sadly, this is not always implemented in the
API of libraries. For example, NumPy's unfunc do not have a `__qualname__`, as
discussed in https://github.com/numpy/numpy/issues/4952. I noticed that other
parts of the NeuroLang code **do check** for the presence of the `__qualname__`
attribute before accessing it. For example, in the `Expression` class

```python
    @property
    def __type_repr__(self):
        if (
            hasattr(self.type, '__qualname__') and
            not hasattr(self.type, '__args__')
        ):
            return self.type.__qualname__
        else:
            return repr(self.type)
```

### DIFF
--- a/neurolang/expressions.py
+++ b/neurolang/expressions.py
@@ -548,7 +548,11 @@ class Constant(Expression):
     def __repr__(self):
         if self.value is ...:
             value_str = '...'
-        elif callable(self.value) and not isinstance(self.value, Expression):
+        elif (
+            callable(self.value)
+            and not isinstance(self.value, Expression)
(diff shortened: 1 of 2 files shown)


demianw:
Good catch. Funny that all python 3.4 all functions should have it and python produces automatically. 🤷 https://www.python.org/dev/peps/pep-3155/
 Will merge when tests pass

PR #581, opened 2020-11-24
Initialise frontend sets from pandas dataframes

demianw:


examples/plot_neurosynth_implementation.py
```diff
@@ -90,18 +90,17 @@ def agg_create_region_overlay(
...
+docs = nl.add_uniform_probabilistic_choice_over_set(ns_docs, name="docs")
 
 
+# %%
 ###############################################################################
 # Probabilistic program and querying
 
 with nl.scope as e:
-    e.vox_activation[e.i, e.j, e.k, e.d] = e.activations[
+    e.vox_activation[e.i, e.j, e.k, e.d] = e.activations(
```
tgy, line 103: why this change of syntax? just wondering

examples/plot_neurosynth_implementation.py
```diff
@@ -90,18 +90,17 @@ def agg_create_region_overlay(
...
 # Loading the database
 
-activations = nl.add_tuple_set(ns_database.values, name="activations")
-terms = nl.add_tuple_set(ns_terms.values, name="terms")
-docs = nl.add_uniform_probabilistic_choice_over_set(
-    ns_docs.values, name="docs"
-)
+activations = nl.add_tuple_set(ns_database, name="activations")
+terms = nl.add_tuple_set(ns_terms, name="terms")
+docs = nl.add_uniform_probabilistic_choice_over_set(ns_docs, name="docs")
```
tgy, line 95: this looks much cleaner than having to suffix a `.values`!

tgy (approved):
Merged this with my example's branch and can confirm this works! Thanks

examples/plot_neurosynth_implementation.py
demianw: No, both work. I think it's nicer if it's `()` on the body of the rules but open to discussion

examples/plot_neurosynth_implementation.py
demianw: Yup! Thanks for pushing for it!

examples/plot_neurosynth_implementation.py
tgy: alright, that works for me. will adapt other examples to use this syntax

PR #584, opened 2020-12-03
Hashing of RA set with None _container

tgy:


neurolang/utils/relational_algebra_set/pandas.py
```diff
@@ -342,6 +342,8 @@ def itervalues(self):
             return iter(self._container.itertuples(name=None, index=False))
 
     def __hash__(self):
+        if self._container is None:
+            return hash(None)
```
tgy, line 346: not sure about what should be returned here but the following code assumes that `self._container` is a proper `pd.DataFrame`, which is not the case if it's `None` of course.

demianw: Good catch with the corner case. Is there a reason why the hash should be the same than that of None? I would have gone for:
```
hash(tuple(), None)
```
to keep some consistency.

In terms of the testing, you just need to test that this hash doesn't explode. Testing the consistency of the hash function is a bit more complex and maybe out of scope.

demianw (changes requested):

PR #585, opened 2020-12-03
Spatial prior through query-based probabilistic facts

tgy:


neurolang/datalog/expression_processing.py
```diff
@@ -655,7 +655,7 @@ class _RuleNormaliser(
         pass
 
     @add_match(
-        FunctionApplication,
+        FunctionApplication(Symbol, ...),
```
tgy, line 658: more restrictive pattern to be able to catch builtin function applications in other pattern

neurolang/datalog/expression_processing.py
```diff
@@ -926,3 +932,41 @@ def implication_with_collapsable_conjunctive_antecedent(self, implication):
                 implication.consequent, self.walk(implication.antecedent)
             )
         )
+
+
+def is_to_be_sorted_equality(formula):
+    return (
```
tgy, line 938: used to detect equalities where the lhs is not the symbol (e.g. `EUCLIDEAN(...) = d`)

neurolang/datalog/wrapped_collections.py
```diff
@@ -154,9 +154,13 @@ def is_constant_tuple_or_tuple_of_constants(val):
         )
 
     def __eq__(self, other):
+        if not isinstance(other, Iterable):
+            return False
```
tgy, line 158: we added this during our pair programming session @demianw

neurolang/utils/relational_algebra_set/pandas.py
```diff
@@ -851,6 +851,8 @@ def __sub__(self, other):
             if other.is_dee():
                 return self.dum()
             return self.dee()
+        self._drop_duplicates_if_needed()
+        other._drop_duplicates_if_needed()
```
tgy, line 855: we added this to optimise the code

neurolang/frontend/datalog/sugar/__init__.py
```diff
@@ -332,12 +332,8 @@ class TranslateSSugarToSelectByColumn(ew.PatternWalker):
 
     @ew.add_match(Implication, _RECOGNISE_S_SUGAR.walk)
     def replace_s_getattr_by_first_column(self, expression):
-        new_expression = (
-            TranslateSSugarToSelectByColumn
-            ._convert_attr_s__to_SelectByColumn
-            .walk(
-                expression
-            )
+        new_expression = TranslateSSugarToSelectByColumn._convert_attr_s__to_SelectByColumn.walk(
```
demianw, line 335: Less clear code due to very opinionated formatting, please roll-back

neurolang/frontend/datalog/sugar/spatial.py
```diff
@@ -0,0 +1,232 @@
...
+
+class DetectEuclideanDistanceBoundMatrix(PatternWalker):
+    """
+    Detect a Euclidean spatial bound in the antecedent of a rule.
+
+    This spatial bound is defined by 4 conjuncts in the antecedent of the rule:
+
+        - a formula of the form `d = EUCLIDEAN(i1, j1, k1, i2, j2, k2)`, where
+          EUCLIDEAN a special reserved symbol corresponding
+        - a formula of the form `d < MAX_DIST`, where MAX_DIST is a constant
```
demianw, line 32: Could you capture cases where "d == EUCLIDEAN" is present but there is no bound `d < MAX_DIST` and throw an error? I think it's a safer way to implement the sugar case without a user making mistakes until we figure out how to implement more general patterns.

neurolang/frontend/datalog/sugar/tests/test_spatial.py
```diff
@@ -0,0 +1,31 @@
...
+    d = Symbol("d")
+    max_dist = Constant(2)
+    detector = DetectEuclideanDistanceBoundMatrix()
+    expression = Implication(
+        P(i2, j2, k2),
+        Conjunction(
+            (
+                FirstRange(i1, j1, k1),
+                SecondRange(i2, j2, k2),
+                EQ(d, EUCLIDEAN(i1, j1, k1, i2, j2, k2)),
```
demianw, line 26: I think we need to add tests for `EQ(EUCLIDEAN(..), d)` and for `GT(max_dist, d)` both which should work or throw an error. And tests for throwing an error on all cases which are not yet captured by the sugar

neurolang/probabilistic/query_resolution.py
```diff
@@ -58,15 +60,18 @@ def union_with_query_based_pfact(self, union):
                 new_formulas.append(formula)
         return self.walk(Union(tuple(new_formulas)))
 
-    @add_match(Implication, is_query_based_probfact)
+    @add_match(
+        Implication,
+        lambda implication: is_query_based_probfact(implication)
+        and not hasattr(implication, "__already_walked__"),
```
demianw, line 66: This is a bit dirty, are we sure there is no other way to guarantee that we are no walking it twice? This is type of patterns that solve something now but generate inconsistent coding systems later and hence bugs which are impossible to catch.

neurolang/probabilistic/query_resolution.py
```diff
@@ -39,7 +39,9 @@ class QueryBasedProbFactToDetRule(PatternWalker):
     @add_match(
         Union,
         lambda union: any(
-            is_query_based_probfact(formula) for formula in union.formulas
+            is_query_based_probfact(formula)
+            and not hasattr(formula, "__already_walked__")
```
demianw, line 43: This is a bit dirty, are we sure there is no other way to guarantee that we are no walking it twice? This is type of patterns that solve something now but generate inconsistent coding systems later and hence bugs which are impossible to catch.

neurolang/utils/relational_algebra_set/pandas.py
demianw: This should first be verified that it doesn't make the rest of the cases slower. Can you please use [asv](https://asv.readthedocs.io/en/stable/) to run the chase benchmarks and compare this and the master to see that it doesn't make us suffer on other cases?

examples/plot_cbma_spatial_prior.py
```diff
@@ -0,0 +1,143 @@
...
+import nibabel
+import nilearn.datasets
+import nilearn.image
+import nilearn.plotting
+import numpy as np
+import pandas as pd
+
+from neurolang.frontend import ExplicitVBR, ExplicitVBROverlay, NeurolangPDL
+
+###############################################################################
```
demianw, line 24: MINOR: add a `# %%` above each title if you can it will be great to translate them into notebooks making them compatible with [jupytext percent style](https://jupytext.readthedocs.io/en/latest/formats.html#the-percent-format) .

demianw (changes requested):
Very good job. The example reads beautifully. As this PR spills across different changes and modules there are changes in each of them.

examples/plot_cbma_spatial_prior.py
tgy: good idea, i'll do that

neurolang/probabilistic/query_resolution.py
tgy: I fully agree with you. This is dirty. I would love a better solution than this one. I could not find any other solution to prevent infinite recursion. This comes back to handling patterns that match a single rule but that can then generate multiple rules. You mentioned that @Jbat1Jumper implemented something that handles that well in the logic module? I could not find anything that does that. Do you remember more precisely where it was? Alternatively, do you have a cleaner solution that this one? I could not come up with something better yet.

neurolang/frontend/datalog/sugar/tests/test_spatial.py
tgy: i will add those

neurolang/probabilistic/query_resolution.py
tgy: I improved the "tag"'s name a little bit, and also added a `TODO` comment

neurolang/frontend/datalog/sugar/tests/test_spatial.py
tgy: done

neurolang/frontend/datalog/sugar/spatial.py
tgy: done

neurolang/frontend/datalog/sugar/__init__.py
tgy: done

neurolang/probabilistic/query_resolution.py
demianw: OK, great I will see if I can figure out a better way to do this (I'll have to understand the essence of the algorithm first feel free to help, but low priority on your list)

examples/plot_cbma_spatial_prior.py
```diff
@@ -21,25 +26,29 @@
 
 from neurolang.frontend import ExplicitVBR, ExplicitVBROverlay, NeurolangPDL
 
-###############################################################################
+# %% [markdown]
```
demianw, line 29: Careful, I was suggesting just adding the "# %%" on top for a reason (https://sphinx-gallery.github.io/stable/tutorials/plot_parse.html ) to keep the notebook bi-compatible between jupytext and sphinx-gallery

tgy:
@demianw I've re-ran the benchmark for the current branch. the drop duplicate function is now called only for natural join and set substraction

the output of the benchmark now says `BENCHMARKS NOT SIGNIFICANTLY CHANGED.` for `asv continuous -v $(git rev-parse origin/master) $(git rev-parse HEAD) --environment=conda:3.7` on the `drago` machine.

neurolang/probabilistic/query_resolution.py
tgy: @demianw did you see the docstring just above? i documented the algorithm in the class. i think this was my best attempt at explaining what the algorithm is doing. let me know if i can help explaining this further

neurolang/probabilistic/query_resolution.py
```diff
@@ -39,7 +39,9 @@ class QueryBasedProbFactToDetRule(PatternWalker):
     @add_match(
         Union,
         lambda union: any(
-            is_query_based_probfact(formula) for formula in union.formulas
+            is_query_based_probfact(formula)
+            and not hasattr(formula, "__QueryBasedProbFactToDetRule__")
```
demianw, line 43: Maybe we could check here just that it's a `is_query_based_probfact(formula)` and the body requires the rewriting step? I'm guessing that checking that the body is a single predicate and `_f2_` is a variable on such predicate would suffice no? Then the only probfact required rewriting are the ones that are not like that. This will eliminate some trivial rewritings *and* the need for the new attribute. What do you think @tgy ?

demianw (changes requested):
Just one comment @tgy let me know what you think about it.

neurolang/probabilistic/query_resolution.py
demianw: @tgy we will need to solve this before merging. Mainly because it's leaving/based on underspecified attributes in objects and that breeds for poor maintainability

tgy: see last commit

PR #589, opened 2021-01-06
Fix codacy coverage reporting

demianw:


demianw:
@jonasrenault I had to fix the pandas version to be lower than 1.2.0 as there is an issue with the new version. We should look into this

PR #593, opened 2021-01-18
Prevent some post-probabilistic rule to be considered as probabilistic

tgy:
Fix an error where a rule depending on a post-probabilistic rule was considered
to be probabilistic, which cannot happen as any rule depending on a
post-probabilistic rule must be deterministic and in the post-probabilistic
stratum.

### DIFF
--- a/neurolang/probabilistic/stratification.py
+++ b/neurolang/probabilistic/stratification.py
@@ -175,7 +175,7 @@ def _get_rule_idb_type(rule, grpd_symbs, wlq_symbs):
         | grpd_symbs["post_probabilistic"]
     ).issuperset(dep_symbs):
         idb_type = "post_probabilistic"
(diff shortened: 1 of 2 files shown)


neurolang/probabilistic/stratification.py
```diff
@@ -175,7 +175,7 @@ def _get_rule_idb_type(rule, grpd_symbs, wlq_symbs):
         | grpd_symbs["post_probabilistic"]
     ).issuperset(dep_symbs):
         idb_type = "post_probabilistic"
-    elif not grpd_symbs["probabilistic"].isdisjoint(dep_symbs):
+    elif not (grpd_symbs["probabilistic"] - wlq_symbs).isdisjoint(dep_symbs):
```
tgy, line 178: this is a simple fix that excludes relations defined by within-language queries from the dependency check.

a rule is probabilistic here if it depends on another probabilistic rule that is NOT a WLQ

PR #596, opened 2021-01-27
Make the CMBA example run in MNI coordinates

demianw:


examples/plot_cbma_spatial_prior.py
```diff
@@ -116,21 +108,21 @@ def agg_create_region_overlay(
 nl.add_symbol(np.exp, name="exp", type_=Callable[[float], float])
 
 with nl.environment as e:
```
demianw, line 110:
```suggestion
with nl.scope as e:
```

tgy (approved):
lgtm

PR #600, opened 2021-02-01
Improve error message when projection on unknown col

tgy:


neurolang/datalog/chase/relational_algebra.py
```diff
@@ -189,10 +189,10 @@ def chase_step(self, instance, rule, restriction_instance=None):
             return self.compute_result_set(
                 rule, substitutions, instance, restriction_instance
             )
-        except ProjectionOverMissingColumnsError:
+        except ProjectionOverMissingColumnsError as e:
             raise WrongArgumentsInPredicateError(
-                "There is a preducate in the query with the "
-                "wrong number of arguments."
+                "There is a predicate in the query with the "
+                "wrong number of arguments. {}".format(str(e))
```
demianw, line 195: I think that the best way to enrich the exception message such that it could be caught and be better expressed at a higher level like the frontend is to add exception data, such as the predicate name, as parameters on the exception class initialiser. Check for instance https://docs.python.org/3/library/exceptions.html#OSError

demianw:
@tgy just curious, has this problem been solved?

tgy:
Yes it seems that the `master` branch has a proper error message now.

See here: https://github.com/NeuroLang/NeuroLang/blob/master/neurolang/relational_algebra.py#L653

You can close this issue, I believe.

PR #602, opened 2021-02-03
[smallfix] only unify symb-to-symb or symb-to-const equality

tgy:
this is currently failing for expressions with builtin function applications


demianw (changes requested):
Good catch! Maybe add a test?

PR #605, opened 2021-02-11
Add test for nl.exists (segregation queries with negated existentially quantified sub-part of the body)

tgy:
I wonder if there is a problem with the implementation
of the negation for existentially quantified variables

We stumbled upon the issue when trying to formulate
queries that segregate more than one network.

Something like the following

```python
e.StudyMatchingNetworkSegregationQuery[e.s, e.n] = (
    Network(e.n2)
    & (e.n2 != e.n)
    & e.NetworkReported(e.n, e.s)
    & ~e.NetworkReported(e.n2, e.s)
)
```

I added a test that fails, which I believe reproduces
this issue.

### DIFF
--- a/neurolang/frontend/tests/test_negation.py
+++ b/neurolang/frontend/tests/test_negation.py
@@ -42,7 +42,7 @@ def assert_almost_equal(set_a, set_b):
         )
 
 
-def test_postprob_conjunct_with_wlq_result():
+def test_deterministic_segregation_query():
     nl = NeurolangPDL()
     nl.add_tuple_set(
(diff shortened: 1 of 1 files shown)


tgy:
This was a problem of understanding the logic semantic of the rule. It should be written as
```python
        e.SegregationQuery[e.n, e.s] = e.NetworkReported(
            e.n, e.s
        ) & ~nl.exists(e.n2, e.NetworkReported(e.n2, e.s) & (e.n2 != e.n))
```
because the existential quantifier should only apply to that part of the body (by default it applies to the whole body)

PR #611, opened 2021-02-26
Handle tuple unicity and aggregation in query-based probabilistic facts

tgy:


neurolang/frontend/probabilistic_frontend.py
```diff
@@ -146,6 +146,9 @@ def __init__(
         self.probabilistic_solvers = probabilistic_solvers
         self.probabilistic_marg_solvers = probabilistic_marg_solvers
         self.ontology_loaded = False
+        self.symbol_table[ir.Symbol("MAX")] = ir.Constant[
+            Callable[[Iterable], Any]
+        ](max)
```
tgy, line 151: a bit ugly, i'm not sure where to put these. a mixin?

neurolang/probabilistic/query_resolution.py
```diff
@@ -16,6 +17,9 @@
 )
 from .expressions import Condition, ProbabilisticPredicate
 
+AGG_MEAN = Symbol("MEAN")
+AGG_MAX = Symbol("MAX")
```
tgy, line 21: where should these be defined?

neurolang/probabilistic/query_resolution.py
```diff
@@ -186,9 +214,26 @@ def _add_to_probabilistic_program(add_fun, pred_symb, expr, det_edb):
...
+    if length_without_probs != length:
+        n_repeated_tuples = length - length_without_probs
+        raise MalformedProbabilisticTupleError(
+            "Some tuples have multiple probability labels. "
+            f"Found {n_repeated_tuples} tuple repetitions, out of "
+            f"{length} total tuples. If your query-based probabilistic fact "
+            "leads to multiple probabilities for the same tuple, you might "
+            "want to consider aggregating these probabilities by taking their "
+            "maximum or average."
+        )
```
tgy, line 234: not sure about the best error message. this is a proposal

neurolang/frontend/probabilistic_frontend.py
demianw: yes, the `TypedSymbolTableMixin` has a way to add functions in mixins (see `included_functions`) which but this was not implemented for symbols

neurolang/probabilistic/query_resolution.py
demianw: It's OK. Do you want to add the `n_repeated_tuples` and `length` or the name of the rule as parameters to the Error such that we can catch and rethrow later with another message, if needed, at the fronted level? I think we should prepare things to not throw any IR exception from the frontend and make it easy for the front-end to catch and re-throw the errors which can be specific to each different frontend

neurolang/probabilistic/query_resolution.py
demianw: If we are adding default aggregations maybe in the datalog aggregation module?

neurolang/datalog/aggregation.py
```diff
@@ -40,6 +43,22 @@
...
+AGG_MEAN = Symbol("MEAN")
+
+
+def is_builtin_aggregation_functor(functor):
+    return functor in (AGG_MAX, AGG_MEAN)
+
+
+class BuiltinAggregationMixin:
+    def function_MAX(self, iterable: typing.AbstractSet) -> Unknown:
+        return numpy.max(iterable)
```
demianw, line 56: It's not a good idea to wrap them this way, as for instance Pandas will detect or use the `np.max` or `np.mean` functions and used optimised versions. You optimised for cases like this in the RA solver module.

tgy: do you mean to use Python's builtin `max` function? what about the `mean` aggregation? there is no builtin for this one

demianw: I mean providing a symbol which points to a constant with `np.max` or `np.mean` instead of a wrapping function.

demianw: same way you did with `np.exp`

neurolang/probabilistic/query_resolution.py
demianw: @tgy on a second note these checks should be `optional`, as they could have a great computational cost. We should discuss with @jonasrenault on maybe making a singleton class which holds all configuration settings for NeuroLang (as he does for the RA backend) such that these options could be centralised.

neurolang/expressions.py
```diff
@@ -859,6 +860,21 @@ def add_functions_to_symbol_table(self):
             top_scope = top_scope.enclosing_scope
         top_scope.enclosing_scope = keyword_symbol_table
 
+    def add_included_constant_functions_to_symbol_table(self) -> None:
```
demianw, line 863: this is better as a `constant_keyword_symbols` no? Nothing in what you did is specific to functions but just to symbols pointing to constants. Can you rename it everywhere else?
Also the use of a dictionary with a specific name means that the mixin strategy will fail if there are more than two mixins defining symbols. Why didn't you just copy the strategy with `included_functions` by adding constants, something like
```python
class MyNewMixin:
    constant_exp = Constant(np.exp)
```
 or something like that? Then you just have to copy __exactly__ the way it's done with functions but with constants.

neurolang/relational_algebra.py
```diff
@@ -453,6 +453,36 @@ def numpy_exponential(self, fa):
...
+        )
+
+    @ew.add_match(
+        FunctionApplication(Constant, ...),
+        lambda fa: (
+            isinstance(fa.functor.value, Callable)
+            and numpy.mean == fa.functor.value
+        ),
+    )
+    def numpy_mean(self, fa):
```
demianw, line 478: maybe we could just make a dictionary with all these and implement it as cleanly as with the `OPERATOR_STRING` case for the arithmetic operators?

demianw (changes requested):
Good advancement, some things to change still

neurolang/expressions.py
tgy: sounds good, did not understand that this was what you had in mind. now i get it

neurolang/relational_algebra.py
tgy: yes, i thought of doing it but as i was pressured by time i went for the path of least resistance. i'll do that

neurolang/relational_algebra.py
```diff
@@ -394,6 +394,22 @@ def __init__(self, relation, column_name, column_value):
     operator.pow: "**",
 }
 
+TRANSLATABLE_NUMPY_OPERATION = {
```
demianw, line 397: this will work with any operation no? It doesn't have to be numpy. In fact python's built-ins `max`, `min`, `round`, `pow` and `math` module `log`, `sqrt`, `exp`, etc could be easily added too now. All being translated to the corresponding ANSI SQL or  PANDAS/Numexpr functions which are listed here https://numexpr.readthedocs.io/projects/NumExpr3/en/latest/user_guide.html#supported-functions https://pandas.pydata.org/pandas-docs/stable/reference/groupby.html#computations-descriptive-stats

you can take 10 minutes to make the full list and then you are done instead of making a PR per new operation that you need.

this of course deprecates #612

demianw (changes requested):
Almost there!

neurolang/datalog/expression_processing.py
```diff
@@ -959,6 +953,29 @@ def add_equality_with_symbol(eq_sets, first, second):
...
+            if isinstance(old_arg, FunctionApplication) and any(
+                isinstance(a1, Symbol) and isinstance(a2, Constant)
+                for a1, a2 in zip(old_arg.args, new_arg.args)
+            ):
+                problematic_replacements = set(
+                    (a1, a2)
+                    for a1, a2 in zip(old_arg.args, new_arg.args)
+                    if isinstance(a1, Symbol) and isinstance(a2, Constant)
+                )
+                raise ForbiddenExpressionError(
```
demianw, line 968: Can you make a more specific exception class? This will help having better error reporting in the frontend.

neurolang/relational_algebra.py
```diff
@@ -394,6 +396,50 @@ def __init__(self, relation, column_name, column_value):
...
+    "arcsin": "asin",
+    "arccos": "acos",
+    "arctan": "atan",
+    "arcsinh": "asinh",
+    "arccosh": "acosh",
+    "arctanh": "atanh",
+}
+EVAL_OP_TO_STR.update(
+    {
+        getattr(
```
demianw, line 424: this is very complex i.e. low maintenability code. Can you make proper loops and ifs and a function and then just calls
`EVAL_OP_TO_STR = f()` ?

demianw (approved):
Two small changes and it's in.

neurolang/relational_algebra.py
```diff
@@ -396,41 +396,54 @@ def __init__(self, relation, column_name, column_value):
...
+        "arcsin": "asin",
+        "arccos": "acos",
+        "arctan": "atan",
+        "arcsinh": "asinh",
+        "arccosh": "acosh",
+        "arctanh": "atanh",
     }
-)
+    eval_op_to_str.update(
+        {
```
demianw, line 432: and you still kept the super complex and unmaintainable code with the dictionary defined with inline ifs.

demianw: That's what I meant by `Can you make proper loops and ifs`

neurolang/relational_algebra.py
```diff
@@ -440,30 +488,15 @@ def arithmetic_operation(self, fa):
...
     @ew.add_match(
         FunctionApplication(Constant, ...),
-        lambda fa: (
-            isinstance(fa.functor.value, Callable)
-            and numpy.exp == fa.functor.value
-        ),
+        is_translatable_numpy_operation,
     )
-    def numpy_exponential(self, fa):
+    def translatable_numpy_operation(self, fa):
```
demianw, line 493: This is not numpy specific anymore, maybe `translatable_operation`

neurolang/datalog/expression_processing.py
```diff
@@ -959,6 +954,29 @@ def add_equality_with_symbol(eq_sets, first, second):
         if not found_eq_set:
             eq_sets.append({first, second})
 
+    @staticmethod
+    def _check_no_agg_var_replaced_by_const(old_csqt, new_csqt) -> None:
+        for old_arg, new_arg in zip(old_csqt.args, new_csqt.args):
+            if isinstance(old_arg, FunctionApplication) and any(
```
demianw, line 960: If you want to check for aggregation, there is a specific class to distinguish aggregations from functions in the head: `AggregationApplication`

demianw (approved):
Two small details and then merge

neurolang/datalog/expression_processing.py
tgy: I've updated the code to check for an `AggregationApplication` instead. I think this is what you meant but I have a small doubt. Confirmed?

neurolang/relational_algebra.py
tgy: Done!

neurolang/datalog/expression_processing.py
```diff
@@ -394,7 +394,7 @@ def dependency_matrix(datalog, rules=None, instance=None):
     ------
     SymbolNotFoundError
         If there is a predicate in the antecedent of a rule which
-        is not a constant or an extension/intensional predicate
+        is not a constant or an extensional/intensional predicate.
```
tgy, line 397: Just a small typo that was wrongly fixed by a recent commit from some open source contributor

PR #623, opened 2021-05-27
Parser/Xrewriter Optimization

gzanitti:
I made some modifications to the Parser and the Xrewrite algorithm to optimize queries that include ontological information.

Among the design decisions involved:
- The parser now returns a dictionary of constraints organized according to the functor of the consequent of the rules. Xrewriter now takes this input, allowing to minimize the number of iterations.
- Now the parser also returns a set of rules, organized in the same way as the constraints, and a new operation `_add_bulk_rules` allows to enter them in the program base without using the walk operation.

EDIT: `_add_bulk_rules` has been removed because it's not needed in this PR. It will be added in a new PR to simplify the review of this one.

____________
Summary of changes by file:

- `neurolang/datalog/chase/relational_algebra.py`: Avoid removing computed operations on empty sets (we work together with Demian to solve this).
- `neurolang/datalog/constraints_representation.py`:
    - `__constraints__` is now a symbol, at Jonas' request.
    - code adapted to be able to receive/return the dictionary of terms required by the new XRewriter
    - New method `add_existential_rules` that returns terms with existential in the head. (See `expression_processing.py` for more details)
- `neurolang/datalog/expression_processing.py`: Due to the use of support expressions to construct existentials (see: https://github.com/NeuroLang/NeuroLang/blob/fac45c93f3dd8ba0a3cb2ff731f547e0d0786e8a/neurolang/datalog/ontologies_parser.py#L321), it is necessary that the consequent of these atoms to be taken as a symbol (but they are not in the `symbol_table`!), so checks are added to take them into account. 
- `neurolang/datalog/ontologies_parser.py`: New parser
- `neurolang/datalog/ontologies_rewriter.py`: Optimizations to avoid going through the entire list of constraints
- `neurolang/frontend/probabilistic_frontend.py`: Adapted to use the new parser
- `neurolang/probabilistic/cplogic/program.py`: Symbols derived from the ontology that do not have values in the symbol table are represented as constants. Code to check this situation has been added.
- `neurolang/probabilistic/stratification.py`: same as `expression_processing.py`

- `neurolang/frontend/tests/test_probabilistic_frontend.py`: Older test removed. Idea included in the parser test set.
- `neurolang/datalog/tests/test_ontologies_parser.py`: Older tests adapted to the new parser/rewriter. New tests included
- `neurolang/datalog/tests/test_ontologies_rewriter.py`: Older tests adapted to the new parser/rewriter. New tests included
- `neurolang/probabilistic/tests/test_deterministic_probabilistic_separation.py`: Updated test according to the changes of the parser


neurolang/frontend/probabilistic_frontend.py
```diff
@@ -372,8 +365,8 @@ def _solve(self, query=None):
 
     def _solve_deterministic_stratum(self, det_idb):
         if "__constraints__" in self.symbol_table:
-            eB = self._rewrite_program_with_ontology(det_idb)
-            det_idb = Union(det_idb.formulas + eB.formulas)
+            det_idb = self._rewrite_program_with_ontology(det_idb)
+            self.current_program_rewrited = det_idb
```
demianw, line 369: This might be dangerous what happens if there are different queries executed? Is the rewriting triggered at each point? Is this variable supposedly observable?

neurolang/datalog/basic_representation.py
```diff
@@ -221,6 +221,10 @@ def intensional_database(self):
             )
         }
 
+    def add_bulk_rules(self, symbol, rules):
+        symbol = symbol.cast(UnionOfConjunctiveQueries)
+        self.symbol_table[symbol] = Union(rules)
```
demianw, line 226: Does this add one set of rules for a single symbol? Can you add the documentation?

neurolang/datalog/constraints_representation.py
```diff
@@ -52,6 +53,34 @@ def add_logic_constraint(self, expression):
     def constraints(self):
         return self.symbol_table.get("__constraints__", Union(()))
 
+    def set_constraints(self, categorized_constraints):
```
demianw, line 56: can you add documentation? What does it mean for a constraint to be "categorised"?

demianw (changes requested):
Some reviews to understand code better.

neurolang/frontend/probabilistic_frontend.py
gzanitti: The main idea is to be able to have a variable to observe the rewrite from the frontend. It's executed each time a query is solved (if it involves a rewrite).  Do you think it's not necessary?

neurolang/datalog/basic_representation.py
gzanitti: Yes, exactly. I will add the documentation, thank you!

neurolang/frontend/probabilistic_frontend.py
demianw: Oh! If that's the point it's OK.

gzanitti:
> Some reviews to understand code better.

@demianw I documented most of the functions to make it easier to keep track of the code.

neurolang/frontend/probabilistic_frontend.py
demianw: `rewrited` does not exist in English, it's an irregular verb

demianw:
> > Some reviews to understand code better.
> 
> @demianw I documented most of the functions to make it easier to keep track of the code.

Some of the comments where I asked for documentation are not addressed yes

gzanitti:
@demianw Ready to be reviewed. Looking forward to your comments :)

gzanitti:
- [x] Clean code.
- [x] Documented functions.
- [x] All tests working.
- [x] Description of changes.

It seems to be ready for review

demianw (commented):
First review!

neurolang/datalog/constraints_representation.py
```diff
@@ -105,7 +106,7 @@ def _categorize_constraints(self, sigma):
         '''Function in charge of sorting the constraints in a dictionary
         using the consequent functor as an index.
 
-        This categorization is useful to obtain the constraints in the
+        This indexation is useful to obtain the constraints in the
```
demianw, line 109: "indexing"

setup.cfg
```diff
@@ -41,8 +41,8 @@ install_requires =
   matplotlib
   seaborn
   typing_inspect
-  rdflib<6.0.0
-  sqlalchemy>=1.4.27
+  rdflib
+  sqlalchemy
```
demianw, line 45: sure that we can remove the minimal version for sqlalchemy?

setup.cfg
```diff
@@ -41,8 +41,8 @@ install_requires =
   matplotlib
   seaborn
   typing_inspect
-  rdflib<6.0.0
-  sqlalchemy>=1.4.27
+  rdflib
```
demianw, line 44: won't we need an "rdflib>6.0.0" now?

demianw (changes requested):
Final changes

setup.cfg
gzanitti: That's the way we have it in master. The versioning was added by me trying to solve the problem I had, but it turned out to be related to something else (it was a test).

Also, I limited the version of tatsu, from version 5.0 onwards they don't support python 3.7. I raised the minimum version of pandas, 1.2.0 gives problems in some operations but CircleCI didn't notice it because it always ran on versions higher than 1.3.5.
Finally, I had to restrict the maximum version of pytest to avoid a bug in the tornado tests for the web app.

gzanitti:
I think it's ready to be merged finally. The last changes were only on the setup.cfg to guarantee that we can run against a clean version of python 3.7 without problems (I explained the modifications in a comment, above this one).

PR #615, opened 2021-03-24
Implementation of the full dichotomy theorem

demianw:

PR #632, opened 2021-07-02
Full dichotomy solver negation

demianw:
Add negation to the full dichotomy solver which follows Dalvi/Suciu from #615


neurolang/probabilistic/transforms.py
```diff
@@ -211,10 +211,17 @@ def minimize_component_conjunction(conjunction):
     """
     if not isinstance(conjunction, Conjunction):
         return conjunction
+    positive_formulas = []
+    negative_formulas = []
```
tgy, line 215: Maybe we can check here that each relational symbol appears only positively or only negatively (e.g. by putting the formulas in `dict`s instead of `list`s and then checking that the intersection of their keys are empty)

tgy (commented):
We should make sure we're only allowing queries written in the Unate UCQ language

neurolang/probabilistic/transforms.py
tgy: Another, maybe cleaner possibility, would be to write a mixin that only matches queries written in Unate UCQ and that raises a "no match for pattern" exception if the query is not syntactically valid

demianw: @tgy isn't what you are proposing here? https://github.com/NeuroLang/NeuroLang/pull/632/files#diff-8bdc451fca96217f85ee03712b268c3249a2dc4747fe6fabc13d0e20cb86a4f9R168 ?

This is mostly because the containment rules that are enunciated in the paper don't seem to apply to negative formulas Dalvi-Suciu section 2.6 so I'm only minimising positive atoms.

tgy: Yes, this comment is outdated by the implementation of this function

demianw: As a principle, checking at the same point that you are executing a complex algorithm might be a bad design choice. Yes you put it in the same place, but making a function more complex might bring in unexpected behaviour. And the computational cost of this queries is in terms of query complexity, when we expect that queries are not composed of many atoms

tgy: I fully agree

PR #646, opened 2021-07-08
GroupByAggregation: check for missing group columns

tgy:


neurolang/relational_algebra.py
```diff
@@ -832,6 +832,11 @@ def _compile_function_application_to_sql_fun_exp(self, fun_exp):
     def aggregate(self, agg_op):
         relation = agg_op.relation
         groupby = list(c.value for c in agg_op.groupby)
+        if any(c not in relation.value.columns for c in groupby):
+            missing = set(
+                c for c in groupby if c not in relation.value.columns
+            )
+            raise NeuroLangException(f"Missing group columns: {missing}")
```
demianw, line 839: Maybe we need more precise exception type here. Also, this should never happen when compiling from Datalog. It means a problem in the datalog compiler. Do you have an example of the datalog (probabilistic or deterministic) where this happens?

@tgy The issue here is that we are not validating the operations in this layer. We could, and to be consistent we should add it everywhere. As this might slow down execution, I don't know if it's priority to add it here. I'm more concerned with the Datalog script that triggered this bug.

PR #649, opened 2021-07-09
LQP Disjoint Project for Probabilistic Choices

tgy:


neurolang/frontend/tests/test_probabilistic_frontend.py
```diff
@@ -480,7 +480,8 @@ def mysum(x):
         res = nl.query((e.x, e.s), e.D[e.x, e.s])
 
     assert len(res) == 2
-    assert res.to_unnamed() == {("a", 0.2 * 0.2 + 0.2 * 0.1), ("b", 0.9 * 0.7)}
+    expected = {("a", 0.2 * 0.2 + 0.2 * 0.1), ("b", 0.9 * 0.7)}
+    assert_almost_equal(res.to_unnamed(), expected)
```
tgy, line 484: Small issue to to comparison of resultings probabilities (numerical precision)

neurolang/frontend/tests/test_probabilistic_frontend.py
```diff
@@ -1073,14 +1072,6 @@ def test_no_tuple_unicity_qbased_pfact():
...
         with pytest.raises(RepeatedTuplesInProbabilisticRelationError):
             nl.query((e.x, e.p), e.Query(e.x, e.p))
-    nl = NeurolangPDL(check_qbased_pfact_tuple_unicity=False)
-    nl.add_tuple_set([(0.2, "a"), (0.5, "b"), (0.9, "a")], name="P")
-    with nl.environment as e:
-        (e.Q @ e.p)[e.x] = e.P(e.p, e.x)
-        e.Query[e.x, e.PROB(e.x)] = e.Q(e.x)
-        result = nl.query((e.x, e.p), e.Query(e.x, e.p))
-    expected = {("a", 0.2), ("b", 0.5), ("a", 0.9)}
-    assert_almost_equal(result, expected)
```
tgy, line 1083: Removing this as the probabilistic table is ill-formed (multiple probability labels for the same tuple) so there is no point checking the result

neurolang/probabilistic/cplogic/testing.py
```diff
@@ -80,185 +60,6 @@ def make_prov_set(iterable, columns):
...
-        ulatex = rap_expression_to_latex(uexp, cpl_program, gm)
-        with open(tex_out_path, "w") as f:
-            f.write("\n\\\\\n".join([latex, slatex, ulatex]) + "\n")
-    result = Projection(
-        uexp, tuple(str2columnstr_constant(arg.name) for arg in qpred_args)
-    )
-    solver = RelationalAlgebraProvenanceCountingSolver()
-    result = solver.walk(result)
-    return exp, result
-
```
tgy, line 260: Removing dead code (there is more to remove but we can do this in another PR)

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -146,16 +243,14 @@ def solve_succ_query(query, cpl_program):
                 "Query %s not liftable, algorithm can't be applied",
                 query
             )
-        # project on query's head variables
-        ra_query = _project_on_query_head(ra_query, shattered_query)
```
tgy, line 150: This is actually an unneeded step

neurolang/frontend/tests/test_probabilistic_frontend.py
demianw: Can't you just leave the tests but put them in scopes instead of environments?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -83,6 +94,94 @@
...
+                prov_col,
+            )
+        )
+        relation = ExtendedProjection(relation, proj_list)
+        relation = self.walk(relation)
+        return ProvenanceAlgebraSet(relation.value, prov_col.value)
+
+    @add_match(DisjointProjection)
+    def disjoint_projection(self, proj_op):
+        return self.projection_rap(
```
demianw, line 166: A more correct strategy would be to rewrite the DisjointProjection expression as a Projection expression, and walk again, instead of calling something in a mixin that you are assuming that it would be coded in other classes

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -83,6 +94,106 @@
...
+            prov_col.value,
+        )
+        return res
+
+    @add_match(Projection, is_provenance_operation)
+    def unlabeled_projection(self, proj_op):
+        return self.walk(IndependentProjection.apply(*proj_op.unapply()))
+
+
+class LiftedQueryProcessingSemiringSolver(
```
demianw, line 190: All of this addition is more appropriately included in the same place as I included the `WeightedNaturalJoin`

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -122,10 +233,8 @@ def solve_succ_query(query, cpl_program):
         )
 
     with log_performance(LOG, "Translation and lifted optimisation"):
-        flat_query_body = enforce_conjunction(
-            lift_optimization_for_choice_predicates(
-                flat_query_body, cpl_program
-            )
+        flat_query_body = lift_optimization_for_choice_predicates(
```
demianw, line 236: If the LQP works in the BID case we handle this optimisation shouldn't be needed anymore. Am I right?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -207,10 +314,131 @@ def dalvi_suciu_lift(rule, symbol_table):
...
+        return False, None
+    nonkey_variables = set.union(
+        *(
+            extract_nonkey_variables(atom, symbol_table)
+            for atom in atoms_with_constants_in_all_key_positions
+        )
+    )
+    symbol_table = symbol_table.copy()
+    for atom in atoms_with_constants_in_all_key_positions:
+        assert isinstance(symbol_table[atom.functor], ProbabilisticChoiceSet)
```
demianw, line 364: Asserts should never be in non-testing code. These should be exceptions.

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: I'll move it there

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -630,3 +870,28 @@ def is_atom_a_deterministic_relation(atom, symbol_table):
         symbol_table.get(atom.functor, None),
         DeterministicFactSet
     )
+
+
+def is_atom_a_probabilistic_choice_relation(atom, symbol_table):
+    return isinstance(symbol_table.get(atom.functor), ProbabilisticChoiceSet)
```
demianw, line 876: This indirection is so small and trivial that it makes code more difficult to read. Please try and avoid the overuse of auxiliary functions

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -630,3 +870,28 @@ def is_atom_a_deterministic_relation(atom, symbol_table):
...
+    requirement of having constants in all key positions (it has no key
+    attribute), or
+    (2) if the atom is a tuple-independent relation, all its attributes are key
+    attributes, and so all its terms must be constants for the requirement to
+    be validated.
+
+    """
+    return is_atom_a_probabilistic_choice_relation(atom, symbol_table) or (
+        not is_atom_a_deterministic_relation(atom, symbol_table)
+        and all(isinstance(arg, Constant) for arg in atom.args)
```
demianw, line 896: You'll also need to ensure that the functor is a symbol, isn't that the case? That bug is hidden in your indirection.

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -207,10 +314,131 @@ def dalvi_suciu_lift(rule, symbol_table):
...
+        for f in (first, second, third)
+    )
+    return True, rap.WeightedNaturalJoin(formulas, (1, 1, -1))
+
+
+def extract_nonkey_variables(atom, symbol_table):
+    """
+    Get all variables in the atom that occur on non-key attributes.
+
+    Makes the assumption that the atom is probabilistic.
```
demianw, line 431: This assumption is not needed. If the atom is deterministic or probabilistic fact set the set should be empty.

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: I got inspired by your own function
```python
def is_atom_a_deterministic_relation(atom, symbol_table):
    return isinstance(
        symbol_table.get(atom.functor, None),
        DeterministicFactSet
    )
```
When I remove this, I find the code less trivial to understand. But if you insist, I can remove it and copy paste the content of this function wherever it is called

demianw: You are right, keep it.

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: Let the query be `ans(x, y) :- Q(x, y), P(x), P(y)`, where `Q` is a probabilistic fact relation and where `P` is a probabilistic choice relation. The `lift_optimization_for_choice_predicates` function produces a rewriting of the query as `ans(x, y) :- Q(x, y), P(x), P(y), x = y`, which is then unified as `ans(x, y) :- Q(y, y), P(y)`. This is however not done by the BID extension of the Dalvi/Suciu algorithm, if I am not wrong.

demianw: I'm sure that your optimisation, when applicable, makes for better queries. The issue is, under which cases the optimisation yields a query that produces the right probabilities. Your optimisation works well with CQs, the issue is when to apply it.

neurolang/relational_algebra.py
```diff
@@ -1313,18 +1313,26 @@ def eliminate_trivial_projection(self, expression):
...
+                expression.attributes,
+            )
         )
 
     @ew.add_match(Projection(ExtendedProjection, ...))
     def try_simplify_projection_extended_projection(self, expression):
         new_relation = self.walk(expression.relation)
         if new_relation is not expression.relation:
-            return self.walk(Projection(new_relation, expression.attributes))
+            return self.walk(
```
demianw, line 1348: Why was this change introduced? Was it necessary? Or you just disagreed with my style? Same as above. And if it was not necessary, isn't this just complexifying the PR in an unnecessary case?

neurolang/relational_algebra.py
```diff
@@ -1313,18 +1313,26 @@ def eliminate_trivial_projection(self, expression):
 
     @ew.add_match(
         Projection(Projection, ...),
-        lambda e: set(e.attributes) <= set(e.relation.attributes)
+        lambda e: (
+            type(e) is type(e.relation)
```
demianw, line 1332: isn't this assured by the patter match?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -40,37 +42,35 @@
...
+    ProbabilisticChoiceSet,
     ProbabilisticFactSet,
     generate_probabilistic_symbol_table_for_query,
 )
 from .probabilistic_semiring_solver import ProbSemiringSolver
 from .query_resolution import lift_solve_marg_query
 from .shattering import shatter_easy_probfacts
+from .small_dichotomy_theorem_based_solver import (
+    RAQueryOptimiser,
+    _maybe_reintroduce_head_variables,
```
demianw, line 64: If we need to import it, then it shouldn't start with `_` as it's not an internal function too. Probably change the name to "add_head_constants_to_projected_variables" or something like this

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -147,16 +155,14 @@ def solve_succ_query(query, cpl_program):
...
-        ra_query = _project_on_query_head(ra_query, shattered_query)
         # re-introduce head variables potentially removed by unification
         ra_query = _maybe_reintroduce_head_variables(
             ra_query, flat_query, unified_query
         )
         ra_query = RAQueryOptimiser().walk(ra_query)
 
     with log_performance(LOG, "Run RAP query"):
-        solver = ProbSemiringSolver(symbol_table)
+        solver = LiftedQueryProcessingSemiringSolver(symbol_table)
```
demianw, line 165: I think it's not a semiring anymore as it has to `+`-style operations

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -189,7 +195,7 @@ def dalvi_suciu_lift(rule, symbol_table):
         rule = MakeExistentialsImplicit().walk(rule)
         result = TranslateToNamedRA().walk(rule)
         proj_cols = tuple(Constant(ColumnStr(v.name)) for v in free_vars)
-        return Projection(result, proj_cols)
+        return rap.IndependentProjection(result, proj_cols)
```
demianw, line 198: Is it this projection the appropriate one regardless of wether the rule is made of a query with all-deterministic predicates, a fact set, or a choice set?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -208,10 +214,137 @@ def dalvi_suciu_lift(rule, symbol_table):
...
+            )
+        )
+        if atoms_with_constants_in_all_key_positions:
+            break
+    else:
+        # did not find a CQ with a valid atom, so we cannot apply the rule
+        return False, None
+    first = add_existentials_except(disjunct, free_variables)
+    second = add_existentials_except(
+        Conjunction(tuple(f for f in dnf_query.formulas if f != first)),
```
demianw, line 317: being it a tuple, isn't it easier to do 
```python
tuple(f for f in dnf_query.formulas[1:])
```
?


If you agree, then per convention change `first` and `second` by `head` and `tail`.

Also, are we sure that any split in head and tail will maintain liftability? Or we need to pick this in a smart way?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -208,10 +214,137 @@ def dalvi_suciu_lift(rule, symbol_table):
...
+        free_variables,
+    )
+    third = add_existentials_except(
+        Conjunction((first,) + second.formulas),
```
demianw, line 321: isn't this just `dnf_query.formulas`?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -208,10 +214,137 @@ def dalvi_suciu_lift(rule, symbol_table):
...
             return plan
+        else:
+            has_safe_plan, plan = disjoint_project(rule_dnf, symbol_table)
+            if has_safe_plan:
+                return plan
 
     return NonLiftable(rule)
 
 
+def disjoint_project(rule_dnf, symbol_table):
```
demianw, line 225: Documentation :) ?

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -208,10 +214,137 @@ def dalvi_suciu_lift(rule, symbol_table):
...
+    if len(rule_dnf.formulas) == 1:
+        return disjoint_project_cnf(rule_dnf.formulas[0], symbol_table)
```
demianw, line 227: I think here it is "project_disjunction" and "conjunction" and not disjunctive normal form (dnf) and conjunctive normal form (cnf). You don't really care about the structure of the query inside of the operation, only if it is liftable.

demianw (changes requested):
Good job! First round of comments

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: Based on Suciu et al. it can be **any other UCQ query**, and so they do not mention a "smart way" of splitting, just finding a CQ that has a probabilistic choice

> Second, suppose Q is a UCQ query of the form Q = Q1 ∨ Q' where Q1 is a conjunctive
query having an atom where all key attribute are constants, and Q' is any other UCQ query.
Then we write P (Q) = P (Q1) + P (Q') − P (Q1 ∧ Q' ), then apply the disjoint-project rule
both to Q1 and to Q1 ∧ Q'.

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: In the deterministic case, you are right: the `Projection` is sufficient, although the `IndependentProjection` also works (but going for the `Projection` is faster). In the case of a probabilistic fact table, the `IndependentProjection` is the right projection due to the tuple-independence assumption. In the case of a probabilistic choice, the algorithm should always use a `DisjointProjection` if there are existentially-quantified variables occurring in the probabilistic choice.

demianw: So then, the point of my question, shouldn't this be implemented differently?

neurolang/probabilistic/dalvi_suciu_lift.py
demianw: So then we should implement the strategy we discussed orally on July 20th

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: If you look at the last version of the code, I changed this back to a `Projection`

neurolang/probabilistic/dalvi_suciu_lift.py
tgy: I don't see the point here. You're given a UCQ in the form `Q1 v Q2 v ... v Qn`, where `Q1, ..., Qn` are CQs. It suffices to look for one CQ `Qi` that has a probabilistic choice, and defines `Q'` as the disjunction of all remaining `Qj, j != i`. Maybe I don't understand something, but I don't see what this strategy is going to add.

demianw: Well, what happens if there is more than one which is liftable as a prob-choice, leaving two options `Qi v Q'` and `Qi' v Q''`. Then suppose that `Q'` is not liftable but `Q''` is. Then, unless we can produce an argument ensuring that this case (where there are two decompositions where one is liftable and the other isn't) doesn't exist, we should put in the proper strategy. Do you think this assurance is possible? It might

tgy: That makes sense. I will try to find an example of such query because we need to test this!

demianw: Maybe you can prove that this can't happen, but you need to be sure

demianw:
The test not passing and PR #660 seems to come from our decompositions not respecting section 2.6 in Dalvi Suciu JACM 2012. We need to find a series of transforms that enables us to do that.

demianw:
@tgy why did you close this branch and deleted it?

tgy:
Accident! I was cleaning up some other branches and accidentally deleted this one. This is fixed now

neurolang/probabilistic/small_dichotomy_theorem_based_solver.py
```diff
@@ -234,3 +247,25 @@ def solve_marg_query(rule, cpl):
         set.
     """
     return lift_solve_marg_query(rule, cpl, solve_succ_query)
+
+
+def _get_existential_variables_in_atoms(query, functor_type):
```
demianw, line 253: isn't this just `get_logic_free_variables(shattered_query.antecedent)-set(shattered_query.consequent.args)`?

neurolang/probabilistic/small_dichotomy_theorem_based_solver.py
```diff
@@ -180,6 +180,19 @@ def solve_succ_query(query, cpl_program):
...
+            isinstance(shattered_query.antecedent, Disjunction)
+            or
+            _get_existential_variables_in_atoms(
+                shattered_query, ProbabilisticFactSet
+            )
+            - _get_existential_variables_in_atoms(
+                shattered_query, ProbabilisticChoiceSet
+            )
+        ):
+            raise UnsupportedSolverError(
```
demianw, line 194: But we could solve choice existentials, couldn't we? If not none of the queries in your submitted paper are valid. Can you please explain this to me?

demianw (changes requested):
Some issues with your approach.

neurolang/probabilistic/small_dichotomy_theorem_based_solver.py
tgy: No because it only gets those existentially-quantified variables occurring in atoms matching the `functor_type` argument, which is going to be either `ProbabilisticFactSet` or `ProbabilisticChoiceSet`.

Your proposal gets **all** existientally-quantified variables, independently of the functor type

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -207,10 +220,182 @@ def dalvi_suciu_lift(rule, symbol_table):
...
+            disjunctive_query, symbol_table
+        )
+    )
+    for disjunct in matching_disjuncts:
+        has_safe_plan, plan = _apply_disjoint_project_ucq_rule(
+            disjunctive_query, disjunct, symbol_table
+        )
+        if has_safe_plan:
+            return plan
+    else:
```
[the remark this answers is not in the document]
tgy: The else is not part of the last version of the code. I think you commented an outdated version. I don't see the else on the last version on GitHub or on my local repository

neurolang/probabilistic/small_dichotomy_theorem_based_solver.py
tgy: Yes! Sorry, this error message is wrong and I forgot to update it. To be clear, the case for which we raise this exception is whenever there is an existientally quantified variable that occurs in a probabilistic fact atom, without occurring in a probabilistic choice atom. If it occurs in a probabilistic choice atom then we have mutual exclusivity, but if if does not then we need to do an independent projection with a noisy-or aggregation, and we defer to the full dichotomy solver.

neurolang/probabilistic/tests/test_dalvi_suciu_lift.py
```diff
@@ -10,9 +10,14 @@
...
     Implication,
 )
+from ...relational_algebra import str2columnstr_constant
+from ...relational_algebra_provenance import DisjointProjection
 from ...utils.relational_algebra_set import NamedRelationalAlgebraFrozenSet
 from .. import dalvi_suciu_lift, transforms
-from ..probabilistic_ra_utils import DeterministicFactSet
+from ..probabilistic_ra_utils import (
+    DeterministicFactSet,
+    ProbabilisticChoiceSet,
```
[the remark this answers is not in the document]
tgy: Will add tests for that!

neurolang/probabilistic/small_dichotomy_theorem_based_solver.py
```diff
@@ -253,3 +250,25 @@ def solve_marg_query(rule, cpl):
...
     """
     return lift_solve_marg_query(rule, cpl, solve_succ_query)
+
+
+def _get_existential_variables_in_atoms(query, functor_type):
+    eq_vars = set()
+    head_vars = set(
+        arg for arg in query.consequent.args if isinstance(arg, Symbol)
+    )
+    if isinstance(query.antecedent, Conjunction):
```
demianw, line 260: Simplify the code below with `extract_logic_atoms`

neurolang/probabilistic/tests/test_dalvi_suciu_lift.py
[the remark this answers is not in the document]
demianw: Good! Check that the proper solver is used in the tests.

demianw (changes requested):
Last changes, hopefully

neurolang/relational_algebra_provenance.py
```diff
@@ -110,6 +111,210 @@ def __repr__(self):
...
+            proj_op.attributes,
+            aggregate_functions,
+        )
+        res = ProvenanceAlgebraSet(
+            self.walk(operation).value,
+            prov_col.value,
+        )
+        return res
+
+    @add_match(Union(ProvenanceAlgebraSet, ProvenanceAlgebraSet))
```
demianw, line 282: Is there a guarantee that this is how we should always implement the union? Is this specified in Suciu?

demianw (commented):
Last comment

PR #651, opened 2021-07-13
Fix __repr__ of Conjunction(...) pattern

tgy:


demianw:
In which case do you get a `Conjunction(...)` case printed? This should only happen when it's the Pattern matching case and hence should be in a decorator and not printed.

tgy:
I have this problem when I use `pytest --log-cli-level=debug` in which case the patterns are printed. However the problem seems to be present in all `__repr__` functions that assume the attributes to be iterable. I think the debugging of the pattern matching is broken and would require more fixing. I needed to debug an issue I have.

demianw:
> I have this problem when I use `pytest --log-cli-level=debug` in which case the patterns are printed. However the problem seems to be present in all `__repr__` functions that assume the attributes to be iterable. I think the debugging of the pattern matching is broken and would require more fixing. I needed to debug an issue I have.

Are these all the logic operators or it breaks also for FunctionApplications?

demianw:
Your strategy is the right one, you might just need to apply it, as you said, to all classes with iterable inputs.

tgy:
I have not checked all classes. I know it also happens to the `ExtendedProjection.__repr__`

PR #652, opened 2021-07-13
Nondeterministic weights in `formula_containments`

tgy:


neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -559,10 +553,10 @@ def _formulas_weights(formula_powerset):
...
             for c0, c1 in ((f0, f1), (f1, f0)):
                 if c1 not in tmp_fix_dict_get(f0) and is_contained(c0, c1):
-                    formula_containments[c0].add(c1)
-                    formula_containments[c0] |= (
-                        formula_containments[c1] -
-                        {c0}
+                    formula_containments[c0] = (
+                        tmp_fix_dict_get(c0)
+                        | {c1}
+                        | (tmp_fix_dict_get(c1) - {c0})
```
tgy, line 559: Without this workaround, this does not work. However, the workaround does not work either because the same key ends up being repeated in the dictionary

neurolang/relational_algebra_provenance.py
```diff
@@ -466,67 +466,6 @@ def _apply_provenance_join_operation(
             self.walk(result).value, res_prov_col.value
         )
 
-    @add_match(WeightedNaturalJoin)
-    def prov_weighted_join(self, join_op):
```
tgy, line 470: Moved to the `RelationalAlgebraProvenanceExpressionSemringSolver`, inherited by `ProbSemiringSolver` used for lifted query processing. Otherwise the `WeightedNaturalJoin` operation was never walked.

tgy: Sure! Well if you look at the inheritance of the `ProbSemiringSolver` you can see that there is no rule matching `WeightedNaturalJoin` operations

demianw:
Does the test here pass with #653

tgy:
The nondeterministic behavior is fixed by your proposed changes!

However, the tests cannot pass as the `Union` of provenance sets is currently implemented as a sum of probabilities while it should be a noisy or. This leads to the wrong probabilistic calculation. This is fixed by the LQP Disjoint PR that changes the default projection to be an independent-project (noisy or of probabilities) and therefore the union is fixed as well.

tgy:
I'm closing this PR and moving the test to the LQP disjoint project PR

PR #653, opened 2021-07-15
Workaround for complex Expression as dict key issue

demianw:
There is a code section in https://github.com/NeuroLang/NeuroLang/blob/1e55827573037bea541240654c6ae9e2bbb68e56/neurolang/probabilistic/dalvi_suciu_lift.py#L549
where there is a bug in which complex logic expressions don't work well as keys of a dictionary providing random results in the equality which we can't reproduce outside the dictionary methods (we don't get why).  This is shown in #652. This PR aims at correcting that by using indexed lists instead of a dictionary and to update the equality of `Symbol` and `Constant` instances as that would have been a part of the issue.

### DIFF
--- a/neurolang/probabilistic/dalvi_suciu_lift.py
+++ b/neurolang/probabilistic/dalvi_suciu_lift.py
@@ -8,6 +8,7 @@ from .. import relational_algebra_provenance as rap
 from ..datalog.expression_processing import (
     UnifyVariableEqualities,
     enforce_conjunction,
+    extract_logic_predicates,
     flatten_query,
 )
 from ..datalog.translate_to_named_ra import TranslateToNamedRA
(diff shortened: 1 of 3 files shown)


demianw:
@tgy do you have a test handy to see if this is working now?

demianw:
@gzanitti can you review this PR please? Thanks

gzanitti:
> @gzanitti can you review this PR please? Thanks

sure, but I don't know exactly what I'm reviewing. No description, nothing.
Can I have some context?

demianw:
> > @gzanitti can you review this PR please? Thanks
> 
> sure, but I don't know exactly what I'm reviewing. No description, nothing.
> Can I have some context?

@gzanitti updated the description. Thanks!

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -584,22 +587,17 @@ def mobius_function(formula, formula_containments, known_weights=None):
         known_weights = dict()
```
gzanitti, line 587: Only minor comment, is that, looking at the code I don't see any use case for `known_weights` to be an optional parameter, but maybe it is useful outside the scope of these two functions (I mention it for double checking).

Everything else looks good to me.

gzanitti (approved):

PR #654, opened 2021-07-15
Enable UCQ Shattering

tgy:
- add UCQ shattering test
- update

### DIFF
--- a/neurolang/datalog/expression_processing.py
+++ b/neurolang/datalog/expression_processing.py
@@ -39,7 +39,7 @@ from ..logic import (
     Union,
 )
 from ..logic import expression_processing as elp
-from ..logic.transformations import CollapseConjunctions
+from ..logic.transformations import CollapseConjunctions, GuaranteeConjunction
 from ..logic.unification import most_general_unifier
 from .exceptions import AggregatedVariableReplacedByConstantError
 from .expressions import AggregationApplication, TranslateToLogic
@@ -503,18 +503,8 @@ def is_ground_predicate(predicate):
     return all(isinstance(arg, Constant) for arg in predicate.args)
 
 
-def enforce_conjunction(expression):
-    if isinstance(expression, Conjunction):
-        return expression
-    elif isinstance(expression, (FunctionApplication, Negation)):
-        return Conjunction((expression,))
-    raise ForbiddenExpressionError(
-        "Cannot conjunct expression of type {}".format(type(expression))
-    )
-
-
 def enforce_conjunctive_antecedent(implication):
-    antecedent = enforce_conjunction(implication.antecedent)
+    antecedent = GuaranteeConjunction().walk(implication.antecedent)
     return implication.apply(implication.consequent, antecedent)
 
 
@@ -565,7 +555,7 @@ class HeadConstantToBodyEquality(PatternWalker):
     )
     def implication_with_constant_term_in_head(self, implication):
         body_formulas = list(
-            enforce_conjunction(implication.antecedent).formulas
+            GuaranteeConjunction().walk(implication.antecedent).formulas
         )
         new_consequent_vars = list()
         for term in implication.consequent.args:
--- a/neurolang/datalog/expression_processing.py
+++ b/neurolang/datalog/expression_processing.py
@@ -514,12 +514,8 @@ def enforce_conjunction(expression):
 
 
 def enforce_conjunctive_antecedent(implication):
-    return implication.apply(
-        implication.consequent,
-        remove_conjunction_duplicates(
-            enforce_conjunction(implication.antecedent)
-        ),
-    )
+    antecedent = enforce_conjunction(implication.antecedent)
+    return implication.apply(implication.consequent, antecedent)
 
 
 def maybe_deconjunct_single_pred(expression):
--- a/neurolang/datalog/expression_processing.py
+++ b/neurolang/datalog/expression_processing.py
@@ -12,7 +12,6 @@ from typing import Iterable
 import numpy as np
 
 from ..exceptions import (
-    ForbiddenExpressionError,
     RuleNotFoundError,
     SymbolNotFoundError,
     UnsupportedProgramError,
--- a/neurolang/logic/transformations.py
+++ b/neurolang/logic/transformations.py
@@ -627,3 +627,28 @@ class GuaranteeDisjunction(IdentityWalker):
     @add_match(..., lambda e: not isinstance(e, Disjunction))
     def guarantee_conjunction(self, expression):
         return Disjunction((expression,))
+
+
+def nary_op_has_duplicated_formulas(nary_op: NaryLogicOperator) -> bool:
+    seen = set()
+    for formula in nary_op.formulas:
+        if formula in seen:
+            return True
+        seen.add(formula)
+    return False
+
+
+class RemoveDuplicatedConjunctsDisjuncts(ExpressionWalker):
+    @add_match(Disjunction, nary_op_has_duplicated_formulas)
+    def disjunction(self, disjunction: Disjunction) -> Disjunction:
+        return self.walk(self._nary_op_without_duplicates(disjunction))
+
+    @add_match(Conjunction, nary_op_has_duplicated_formulas)
+    def conjunction(self, conjunction: Conjunction) -> Conjunction:
+        return self.walk(self._nary_op_without_duplicates(conjunction))
+
+    @staticmethod
+    def _nary_op_without_duplicates(
+        nary_op: NaryLogicOperator,
+    ) -> NaryLogicOperator:
+        return nary_op.apply(tuple(set(nary_op.formulas)))
--- a/neurolang/logic/transformations.py
+++ b/neurolang/logic/transformations.py
@@ -638,7 +638,7 @@ def nary_op_has_duplicated_formulas(nary_op: NaryLogicOperator) -> bool:
     return False
 
 
-class RemoveDuplicatedConjunctsDisjuncts(ExpressionWalker):
+class RemoveDuplicatedConjunctsDisjuncts(LogicExpressionWalker):
     @add_match(Disjunction, nary_op_has_duplicated_formulas)
     def disjunction(self, disjunction: Disjunction) -> Disjunction:
         return self.walk(self._nary_op_without_duplicates(disjunction))
--- a/neurolang/probabilistic/shattering.py
+++ b/neurolang/probabilistic/shattering.py
@@ -1,7 +1,7 @@
 import collections
 import itertools
 import operator
-from typing import AbstractSet, Dict, Tuple
+from typing import AbstractSet
 
 from ..datalog.expression_processing import (
     extract_logic_predicates,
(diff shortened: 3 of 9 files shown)


neurolang/probabilistic/shattering.py
```diff
@@ -213,9 +220,40 @@ def shatter_easy_probfacts(query, symbol_table):
         An equivalent conjunctive query without constants.
 
     """
-    query = enforce_conjunctive_antecedent(query)
+    query = RemoveConjunctionDuplicates().walk(query)
```
demianw, line 223: maybe the best would be to convert it to DNF (there is a function for that) before shattering. Then you know it's a disjunction of conjunctions. And moreover, you should make it ignore conjunctive queries with negation. In that way, it runs for all queries, but for some, it does nothing.

demianw (changes requested):

neurolang/probabilistic/shattering.py
```diff
@@ -121,53 +129,58 @@ class EasyQueryShatterer(ExpressionWalker):
...
-            non_prob_columns[i]
-            for i, arg in enumerate(shatter.args)
-            if not isinstance(arg, Constant)
+        pred_symb = shatter.functor.relation
+        cache_key = (pred_symb,)
+        for i, arg in enumerate(shatter.args):
+            if isinstance(arg, Constant):
+                cache_key += (arg,)
+            else:
+                cache_key += (i,)
```
tgy, line 142: The cache key contains the constants and the indices of the non-constant terms (symbols) such that the same key is given to `P(a, x)` and `P(a, y)` but not the same for `P(a, x)` and `P(b, x)`

neurolang/probabilistic/shattering.py
```diff
@@ -213,9 +226,60 @@ def shatter_easy_probfacts(query, symbol_table):
...
+    conjunctive_query: Implication,
+    shatterer: EasyQueryShatterer,
+) -> Implication:
+    tagger = QueryEasyShatteringTagger()
+    tagged_cq = tagger.walk(conjunctive_query)
+    shattered_query = shatterer.walk(tagged_cq)
     _check_shatter_fully_solved(shattered_query)
+    antecedent = shattered_query.antecedent
+    if isinstance(antecedent, Conjunction):
+        antecedent = Conjunction(tuple(set(antecedent.formulas)))
```
tgy, line 282: I have to remove the duplicated conjuncts to handle corner cases. I'm not sure this is the right place to do so. Maybe we should have our algorithms ensure that formulas are not duplicated?

neurolang/probabilistic/shattering.py
```diff
@@ -220,40 +226,60 @@ def shatter_easy_probfacts(query, symbol_table):
...
+        has_negated_atom = False
+        if isinstance(disjunct, Conjunction):
+            disjunct = Conjunction(tuple(set((disjunct.formulas))))
+            has_negated_atom |= any(
+                isinstance(atom, Negation) for atom in disjunct.formulas
+            )
+        elif isinstance(disjunct, Negation):
+            has_negated_atom = True
+        if has_negated_atom:
+            return query
```
demianw, line 242: Seems that you can't shatter a disjunct with a negation. I think this is not true. If you have: 
Q1 v Q2 v Q3 you can shatter them separately, and if one has a negation, then you just avoid shattering that one.

neurolang/datalog/expression_processing.py
```diff
@@ -514,12 +514,8 @@ def enforce_conjunction(expression):
...
 
 def enforce_conjunctive_antecedent(implication):
-    return implication.apply(
-        implication.consequent,
-        remove_conjunction_duplicates(
-            enforce_conjunction(implication.antecedent)
-        ),
-    )
+    antecedent = enforce_conjunction(implication.antecedent)
+    return implication.apply(implication.consequent, antecedent)
```
tgy, line 518: This function should not take care of removing duplicates in the conjunction. This is fixed.

neurolang/logic/transformations.py
```diff
@@ -627,3 +627,28 @@ class GuaranteeDisjunction(IdentityWalker):
...
+
+
+class RemoveDuplicatedConjunctsDisjuncts(ExpressionWalker):
+    @add_match(Disjunction, nary_op_has_duplicated_formulas)
+    def disjunction(self, disjunction: Disjunction) -> Disjunction:
+        return self.walk(self._nary_op_without_duplicates(disjunction))
+
+    @add_match(Conjunction, nary_op_has_duplicated_formulas)
+    def conjunction(self, conjunction: Conjunction) -> Conjunction:
+        return self.walk(self._nary_op_without_duplicates(conjunction))
```
tgy, line 648: Only walking disjunctions and conjunctions because there is a semantic equivalence when duplicates of disjuncts or conjuncts are removed, which is not necessarily the case for all n-ary logic operators

neurolang/probabilistic/shattering.py
```diff
@@ -5,17 +5,29 @@
...
+from .transforms import convert_to_dnf_ucq
 
 EQ = Constant(operator.eq)
 
 
+def query_to_tagged_set_representation(query, symbol_table):
+    new_antecedent = ReplaceExpressionWalker(symbol_table).walk(
+        query.antecedent
+    )
+    return query.apply(query.consequent, new_antecedent)
```
tgy, line 28: Just moved this up to separate functions and walkers

neurolang/probabilistic/shattering.py
```diff
@@ -121,67 +133,84 @@ class EasyQueryShatterer(ExpressionWalker):
     def __init__(self, symbol_table, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.symbol_table = symbol_table
+        self._cached = dict()
+
+    @add_match(Implication(..., FunctionApplication))
+    def implication(self, implication):
+        conjunctive_query = enforce_conjunctive_antecedent(implication)
+        return self.walk(conjunctive_query)
```
tgy, line 141: Rewrite and delegate to the next pattern

neurolang/probabilistic/shattering.py
```diff
@@ -121,67 +133,84 @@ class EasyQueryShatterer(ExpressionWalker):
...
+        disjuncts = list()
+        for conjunctive_query in disjunctive_query.antecedent.formulas:
+            disjuncts.append(self._shatter_conjunction(conjunctive_query))
+        return Implication(
+            disjunctive_query.consequent,
+            Disjunction(tuple(disjuncts)),
+        )
+
+    def _shatter_conjunction(self, conjunction):
+        tagger = QueryEasyShatteringTagger()
```
tgy, line 161: Tag predicates in each conjunction separately

neurolang/probabilistic/shattering.py
```diff
@@ -213,9 +242,13 @@ def shatter_easy_probfacts(query, symbol_table):
...
-    return shattered_query
+    dnf_query_antecedent = convert_to_dnf_ucq(query.antecedent)
+    dnf_query_antecedent = RemoveDuplicatedConjunctsDisjuncts().walk(
+        dnf_query_antecedent
+    )
+    dnf_query = Implication(query.consequent, dnf_query_antecedent)
+    tagged_query = query_to_tagged_set_representation(dnf_query, symbol_table)
+    shatterer = EasyQueryShatterer(symbol_table)
+    shattered = shatterer.walk(tagged_query)
+    shattered = RemoveTrivialOperations().walk(shattered)
```
tgy, line 253: Re-simplify the query if we added trivial conjunctions / disjunctions

neurolang/probabilistic/shattering.py
```diff
@@ -131,6 +135,33 @@ def __init__(self, symbol_table, *args, **kwargs):
...
+    def disjunctive_query(self, disjunctive_query):
+        disjuncts = list()
+        for conjunctive_query in disjunctive_query.antecedent.formulas:
+            disjuncts.append(self._shatter_conjunction(conjunctive_query))
+        return Implication(
+            disjunctive_query.consequent,
+            Disjunction(tuple(disjuncts)),
+        )
+
+    def _shatter_conjunction(self, conjunction):
```
demianw, line 160: I think you might just be able to compose `EasyQueryShatterer` and `QueryEasyShatteringTagger` as I don't think there are overlapping patterns. Am I right?

neurolang/logic/transformations.py
```diff
@@ -627,3 +627,28 @@ class GuaranteeDisjunction(IdentityWalker):
...
+def nary_op_has_duplicated_formulas(nary_op: NaryLogicOperator) -> bool:
+    seen = set()
+    for formula in nary_op.formulas:
+        if formula in seen:
+            return True
+        seen.add(formula)
+    return False
+
+
+class RemoveDuplicatedConjunctsDisjuncts(ExpressionWalker):
```
demianw, line 641: If you make this inherit from `LogicExpressionWalker` it will automatically walk recursively through all subexpressions that don't match those here.

neurolang/probabilistic/shattering.py
demianw: Nice clean code!

demianw (changes requested):
Looks great! Very last extremely trivial comments

neurolang/datalog/expression_processing.py
```diff
@@ -514,12 +514,8 @@ def enforce_conjunction(expression):
 
 
 def enforce_conjunctive_antecedent(implication):
-    return implication.apply(
-        implication.consequent,
-        remove_conjunction_duplicates(
-            enforce_conjunction(implication.antecedent)
-        ),
-    )
+    antecedent = enforce_conjunction(implication.antecedent)
```
demianw, line 517: One more, will `enforce_conjunction` be the same as the existing `GuaranteeConjunction`?

neurolang/probabilistic/shattering.py
demianw: But do we need that? I thought that if we are just shattering constant selections of the shape `P(a, x)` we need to keep track of them across all conjuncts to track dependancies

neurolang/probabilistic/shattering.py
tgy: Done

neurolang/logic/transformations.py
tgy: Done

neurolang/datalog/expression_processing.py
tgy: Done, refactored

neurolang/probabilistic/shattering.py
tgy: True, this should be checked accross conjuncts within the disjunction. And it should fail if there is a dependency between two disjuncts as well. See the test with query `ans(x, y) :- ( P(a, x) ^ Q(x, y) ) v ( Q(y, hello) ^ P(a, x) ^ P(b, y) )` as an example where `Q(x, y)` and `Q(y, hello)` do generate a dependency and therefore the **Union** LQP rule cannot be applied because the query cannot be written as `Q = Q1 v Q2` where `Q1` and `Q2` are independent

tgy: This is now fixed

demianw (approved):
Looks great! Will merge once all tests pass

PR #655, opened 2021-07-15
Ontological structural knowledge

gzanitti:
Incremental PR that works on top of https://github.com/NeuroLang/NeuroLang/pull/623

In addition to the changes introduced in the above mentioned PR, this new PR includes everything necessary to be able to perform queries on the ontology structure and not only on the data. 

All changes introduced in #623 should have been included in this one, so it is recommended to merge the aforementioned PR first.


gzanitti:
Modifications included in this PR:
- Since the way we model entities is based on defining them as atoms, `E.g: Auditory(x, y)`, we need a way to be able to traverse the properties of the ontology, for example, to calculate the synonyms for a term. This PR then, adds symbols of the form `neurolang:property(entity, value)` where the property can be any property included in the ontology, for example `label`. The chosen namespace `neurolang:`, may need to be modified.
- Necessary modifications so that the rules derived from the ontology are not manipulated by the Magic Sets module, avoiding problems when deriving open knowledge.
- The creation of a connector atom (with a user-defined name) at the moment of loading the ontology in neurolang, that allows to connect in a simple way, external data with the entities of the ontology (for example, the NeuroSynth terms).
- Automatically derived rules containing the connector atom mentioned in the previous point are added to the defined program so that it can be used in queries.
- Existentially quantified atoms are included when parsing the program to avoid "Symbol not found" errors since it's possible that some of these variables do not contain entries in the symbol table.

neurolang/datalog/ontologies_parser.py
```diff
@@ -17,13 +17,21 @@ class OntologyParser:
     from an ontology, both at entity and constraint levels.
     """
 
-    def __init__(self, paths, load_format="xml"):
+    STRUCTURAL_KNOWLEDGE_NAMESPACE = 'neurolang:'
```
demianw, line 20: maybe this should be in the config?

Also you should check when you load the ontology that it's not clashing. If not we might have undetectable bugs

neurolang/frontend/probabilistic_frontend.py
```diff
@@ -66,7 +67,10 @@
     QueryBasedProbFactToDetRule,
     compute_probabilistic_solution,
 )
-from ..probabilistic.stratification import stratify_program
+from ..probabilistic.stratification import (
+    _get_list_of_intensional_rules,
```
demianw, line 71: importing `_` methods is not something we should do. If we need this, then the method should be promoted to a method without the `_` and then documented.

demianw (changes requested):
Two small comments

PR #656, opened 2021-07-16
Make type printing optional

demianw:
Little PR making a config option to make type printing at the IR level optional


tgy (approved):
LGTM!

PR #668, opened 2021-08-18
Generalise magic sets for aggregation and negation

demianw:


jonasrenault:
I think this is good for a review @demianw

neurolang/datalog/magic_sets.py
```diff
@@ -68,12 +199,122 @@ def magic_rewrite(query, datalog):
...
+    adorned_query : Expression
+        the adorned query rule
+
+    Returns
+    -------
+    Iterable[Expression]
+        the set of magic initialization rules
+    """
+    magic_init_rules = []
+    for predicate in extract_logic_predicates(adorned_query.antecedent):
```
demianw, line 257: This will iterate you the predicates that are of the form `P(x, y, z, ...)` or not `P(x, y, z, ...)`. What happens if you find a negated predicate? Is it verified before that rules have no negation? Also you might get builtins, i.e., `a == b`. Should you also avoid adding facts for these in the init rules?

Or this is only called with the adorned query which has avoided all these?

neurolang/datalog/magic_sets.py
```diff
@@ -68,12 +199,122 @@ def magic_rewrite(query, datalog):
...
+        for each adorned predicate p in body :
+            add the rule Magic(p) :- Magic(h), ^(body predicates left of p)
+    ```
+    """
+    magic_rules = []
+    for rule in adorned_rules:
+        consequent = rule.consequent
+        magic_head = magic_predicate(consequent)
+        body_predicates = (magic_head,)
+        for predicate in extract_logic_predicates(rule.antecedent):
```
demianw, line 286: Same questions as in line 257. Not all predicates have functors (case of negation)

neurolang/datalog/tests/test_magic_sets.py
```diff
@@ -15,7 +23,10 @@
 
 
 class Datalog(
-    TranslateToLogic,
+    TranslateToLogicWithAggregation,
+    BuiltinAggregationMixin,
+    DatalogWithAggregationMixin,
+    DatalogProgramNegationMixin,
```
demianw, line 29: Maybe add a couple tests to see what happens when Negation and builtins (constant functions as predicates, like `operator.eq`)  are present?

demianw (commented):
Good job @jonasrenault . Some minor changes are needed I think.

jonasrenault:
I've update the code to raise an exception if there exists a negated predicate in the rules, since the algorithm does not (yet) support negations.

I've also changed the way the `query_init_rules` are created to initialise all predicates that have a constant argument and not just those that are in the query rule. This allows to propagate constants that are defined inside a program rule and not just those in the query rule.

I've added tests for these two cases. Tests for builtins already existed.

demianw:
> I've update the code to raise an exception if there exists a negated predicate in the rules, since the algorithm does not (yet) support negations.
> 
> I've also changed the way the `query_init_rules` are created to initialise all predicates that have a constant argument and not just those that are in the query rule. This allows to propagate constants that are defined inside a program rule and not just those in the query rule.
> 
> I've added tests for these two cases. Tests for builtins already existed.

Great!

PR #675, opened 2021-08-27
Enable multiple aggregations

demianw:


demianw:
@jonasrenault I might need help getting multiple aggregation to work with the dasksql  engine.

jonasrenault:
I will look into it

PR #677, opened 2021-08-30
Fix Semantics of Disjoint Project

tgy:
Simply removing the conversion of the ProbabilisticChoiceSet to a
ProbabilisticFactSet suffices.

The following code at line 294 of `dalvi_suciu_lift.py`

```python
conjunctive_query = add_existentials_except(
    query, free_variables | nonkey_variables
)
```

already makes sure that only existential variables that are not projected out
by the disjoint project rule are reintroduced before subsequent calls to
`dalvi_suciu_lift`

### DIFF
--- a/neurolang/probabilistic/dalvi_suciu_lift.py
+++ b/neurolang/probabilistic/dalvi_suciu_lift.py
@@ -285,17 +285,12 @@ def disjoint_project_conjunctive_query(conjunctive_query, symbol_table):
             for atom in atoms_with_constants_in_all_key_positions
         )
     )
-    symbol_table = symbol_table.copy()
     for atom in atoms_with_constants_in_all_key_positions:
         if not isinstance(symbol_table[atom.functor], ProbabilisticChoiceSet):
             raise NeuroLangException(
                 "Any atom with constants in all its key positions should be "
                 "a probabilistic choice atom"
             )
-        symbol_table[atom.functor] = ProbabilisticFactSet(
-            symbol_table[atom.functor].relation,
-            symbol_table[atom.functor].probability_column,
-        )
     conjunctive_query = add_existentials_except(
         query, free_variables | nonkey_variables
     )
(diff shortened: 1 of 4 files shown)


neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -285,17 +285,12 @@ def disjoint_project_conjunctive_query(conjunctive_query, symbol_table):
...
             raise NeuroLangException(
                 "Any atom with constants in all its key positions should be "
                 "a probabilistic choice atom"
             )
-        symbol_table[atom.functor] = ProbabilisticFactSet(
-            symbol_table[atom.functor].relation,
-            symbol_table[atom.functor].probability_column,
-        )
     conjunctive_query = add_existentials_except(
         query, free_variables | nonkey_variables
```
demianw, line 295: what happens if there is a non key variable that has not been projected out and hence, the existential needs to be reinserted? Is this a possibility @tgy ?

neurolang/probabilistic/query_resolution.py
```diff
@@ -200,13 +200,9 @@ def compute_probabilistic_solution(
...
             )
-        else:
-            relation = _solve_for_probabilistic_rule(
-                cpl, rule, succ_prob_solver
+            solution[rule.consequent.functor] = Constant[AbstractSet](
+                relation.value.to_unnamed()
             )
-        solution[rule.consequent.functor] = Constant[AbstractSet](
-            relation.value.to_unnamed()
-        )
```
tgy, line 209: We discussed this small change. This currently leads to queries being computed multiple times and should be removed. The only queries that need to be solved are probobility-encoding rules (previously called within-language probabilistic queries).

tgy: I propose this change here, but I can remove it if you want to do this in another branch.

demianw (approved):

PR #681, opened 2021-09-06
convert_to_union_cnf function

gzanitti:
The new function included in this PR was separated from PR #660 because it is no longer used.


gzanitti:
Hey @demianw, this is the function that was used in the first implementation of Dalvi Suciu, which we then discarded. I separated it to a new PR according to your request, but I see it's been forgotten.

This function is not used anywhere else, and if we include it, it would be dead code.  Should we close this PR?

demianw:
Yup

PR #682, opened 2021-09-07
Fix bug in negation when deterministic sets have constants and self-join

demianw:
Also, made shattering work symbolically


demianw:
Was merged with #683

PR #689, opened 2021-09-15
Optimise pandas extended projection and selection as well as Extended projection static optimizations

demianw:
The main point of this PR is to improve on the static optimiser of RA expressions, with RAP processing in mind. It has

1. Optimisations to the pandas backend to reduce the time processing extended projections
2. Optimizations of the RA expression handling mostly focused on merging extended projections
3. Some minor static optimisation patterns on column renaming
4. Corrected some bugs in adding equiprobable choice facts to the frontend
5. Type inference for natural join operations at the intermediate level
6. Some little code leftovers here and there have been eliminated


tgy (commented):
LGTM. Tried to think of corner cases in which the new optimisation rewriting rules would not apply and could not find any. I would make sure all the new code in the RA module are covered by the tests

PR #700, opened 2021-10-04
Lifted processing of segregation queries

tgy:
There seems to be an issue with the resolution of segregation queries of the
form

```python
e.SegQuery[e.n, e.s] = (
    e.NetworkReported(e.n, e.s)
    & ~nl.exists(
        e.n2,
        (e.n2 != e.n)
        & e.NetworkReported(e.n2, e.s)
    )
)
```

The query is not in the unate fragment of first-order logic. But it is not
detected by the verification step in dalvi/suciu.

Although it is not part of the unate fragment, this query can be shattered
because each ground literal `NetworkReported(e.n, e.s)` in the resulting
conjunction is for a different network `e.n`.

I think we need to have a special case for this type of queries that can be
shattered in this way, which should derive a plan that looks like

```
∏_i{n, s}( σ_{n2 != n}( ρ_{n, s}(NR) ⋈ ρ_{n2, s}(NR) ) )
```

I also saw, while debugging, that the flattening step transforms the query in a
weird way.

It used to work with the WMC resolution but I think now it does not work
anymore.


demianw:
@tgy what's the flattened version of the query?

tgy:
The query
```
Query(n, s) :- SegQuery(n, s), SelectedStudy(s)
```
is flattened as
```
Query(n, s) :- 
    SelectedStudy(s),
    ¬∃ fresh ( NetworkReported(fresh, s), fresh != n, NetworkReported(n, s) ),
    NetworkReported(n, s)
```
but I would have expected something like
```
Query(n, s) :- 
    SelectedStudy(s),
    ¬∃ fresh ( NetworkReported(fresh, s), fresh != n ),
    NetworkReported(n, s)
```

gzanitti:
Hey @tgy, I just fixed a problem in exception inheritance that didn't allow `_solve_probabilistic_stratum` to continue evaluating against the other probabilistic solvers. Your query still fails, but the problem should no longer be in Dalvi Suciu's implementation.

tgy:
I've changed the example to make sure it's safe-range, the resolution seems to be hanging before reaching the probabilistic query evaluation step

demianw:
> I've changed the example to make sure it's safe-range, the resolution seems to be hanging before reaching the probabilistic query evaluation step

@tgy 
Is it hanging or just that the triple cross product generates a horribly large dataset and takes lots of time?
They query was safe range before, just that the way it was rewritten into horn clauses was not what you expected but was it semantically correct?

demianw:
@gzanitti could you check the failing tests? It's just replacing the once from the frontend which are now expected to throw a non-liftable exception, to the fact that we get the right result through WMC

demianw:
@tgy I think we would need a bit of a more formal argument that this is a liftable case. Can you make an example with a set of 3 elements?

PR #727, opened 2022-04-22
Antishattering

gzanitti:


neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -212,6 +225,54 @@ def solve_succ_query(query, cpl_program, run_relational_algebra_solver=True):
...
+        eq_ = Constant(eq)
+        new_selection = False
+        for constant, fresh_var in self.constants_by_formula_dict.items():
+            new_selection = True
+            raoperation = Selection(
+                raoperation,
+                eq_(str2columnstr_constant(fresh_var.name), constant)
+            )
+
+        if len(proyected_variables) > 0 and new_selection:
```
demianw, line 249: Here you should project even if there are not anymore variables if not the case of boolean queries might not work.

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -212,6 +225,54 @@ def solve_succ_query(query, cpl_program, run_relational_algebra_solver=True):
 
     return prob_set_result
 
+class ProjectionSelectionByPChoiceConstant(PatternWalker):
```
demianw, line 228: Maybe we should make an anti-shattering module? And include this in the chained walker at the end of the dalvi-suciu implementation

demianw (commented):
Small comments

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -97,6 +97,10 @@
     minimize_ucq_in_dnf,
     unify_existential_variables
 )
+from .antishattering import (
+    _pchoice_constants_as_head_variables,
+    _selfjoins_in_pchoices
```
demianw, line 102: Avoid using underscore function names for functions exported from the module

neurolang/probabilistic/tests/test_antishattering.py
```diff
@@ -0,0 +1,79 @@
...
+from ..cplogic.program import CPLogicProgram
+
+ans = Symbol("ans")
+P = Symbol("P")
+x = Symbol("x")
+
+a = Constant("a")
+b = Constant("b")
+
+
```
demianw, line 14: Add at least a disjunction case to the tests

neurolang/probabilistic/antishattering.py
```diff
@@ -0,0 +1,75 @@
...
+    generate_probabilistic_symbol_table_for_query,
+    is_atom_a_probabilistic_choice_relation)
+
+
+class ProjectionSelectionByPChoiceConstant(PatternWalker):
+
+    def __init__(self, constants_by_formula_dict):
+        self.constants_by_formula_dict = constants_by_formula_dict
+
+    @add_match(RelationalAlgebraOperation)
```
demianw, line 21: Maybe this should match a ProvenanceAlgebraSet

neurolang/probabilistic/tests/test_antishattering.py
gzanitti: The use of disjunctions causes the query to be non-conjunctive and NeuroLang returns an error. Should I make a test that captures this situation?

demianw: Neurolang shouldn't return an error when the query is non conjunctive. In fact, a segregation query is non conjunctive too.

Can you add the test so we can work on it?

gzanitti: Done. This is where the issue comes from:
https://github.com/NeuroLang/NeuroLang/blob/41a5de6ab421f2f53ed0039cdc257dfc6bb3336e/neurolang/datalog/negation.py#L97

neurolang/probabilistic/tests/test_antishattering.py
```diff
@@ -0,0 +1,96 @@
...
+        (0.52, "a"),
+        (0.48, "b"),
+    }
+
+
+    cpl_program = CPLogicProgram()
+    cpl_program.add_probabilistic_choice_from_tuples(P, table)
+
+    query = Implication(ans(), Disjunction((P(a), P(b))))
+    cpl_program.walk(query)
```
demianw, line 92: The problem with this test is that the `cpl_program` accepts only Datalog-compatible syntax. So you need to have the program

```
A():-P(a)
A():-P(b)
Ans():-A()
```

demianw (changes requested):
Last comments!

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -180,8 +184,13 @@ def solve_succ_query(query, cpl_program, run_relational_algebra_solver=True):
 
     with log_performance(LOG, "Translation to extensional plan"):
         flat_query = Implication(query.consequent, flat_query_body)
+
+        flat_query, constants_by_formula = \
+            pchoice_constants_as_head_variables(flat_query, cpl_program)
+
```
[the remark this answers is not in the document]
gzanitti: The problem with this modification is that I need an instance of the flat_query here:
https://github.com/NeuroLang/NeuroLang/blob/59ea899e5da163240f989c29006ad8993fe6df42/neurolang/probabilistic/dalvi_suciu_lift.py#L210


If I do this modification inside `_prepare_and_optimise_query`, I lose the new variables introduced by `pchoice_constants_as_head_variables `.

If not, it should also return an instance of the query modified by `pchoice_constants_as_head_variables` at the end (in addition to the dictionary of constants to variables created inside), but I think that makes the code more complex.

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -305,6 +314,9 @@ def dalvi_suciu_lift(rule, symbol_table):
         )
 
     rule_dnf = convert_ucq_to_ccq(rule, transformation='DNF')
+    if selfjoins_in_pchoices(rule_dnf, symbol_table):
+        return NonLiftable(rule)
```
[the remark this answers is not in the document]
gzanitti: We decided to add it inside the `inclusion_exclusion_conjunction` method. But I need an instance of the query in dnf form to make it easier to parse. Is it ok if I add it as a condition in the if? 

https://github.com/NeuroLang/NeuroLang/blob/e34136de40fe4a0b04e5d3f0bbfa3683155fd7a5/neurolang/probabilistic/dalvi_suciu_lift.py#L334

It is much simpler to implement and the result should be the same.

neurolang/probabilistic/antishattering.py
```diff
@@ -0,0 +1,300 @@
...
+    NumberColumns,
+    int2columnint_constant,
+    str2columnstr_constant,
+)
+from ..relational_algebra_provenance import ONE, ProvenanceAlgebraSet
+from .probabilistic_ra_utils import is_atom_a_probabilistic_choice_relation
+
+NAMED_DEE = Constant[AbstractSet[Tuple]](
+    WrappedNamedRelationalAlgebraFrozenSet.dee(), verify_type=False
+)
```
demianw, line 38: Maybe move to the same module as `WrappedNamedRelationalAlgebraFrozenSet`

neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -283,33 +295,43 @@ def solve_marg_query(rule, cpl):
...
     [1] Dalvi, N. & Suciu, D. The dichotomy of probabilistic inference
     for unions of conjunctive queries. J. ACM 59, 1–87 (2012).
-    '''
-    rule = RTO.walk(rule)
+    """
+
+    rule = SelfjoinChoiceSimplification(symbol_table).walk(rule)
+    rule = NestedExistentialChoiceSimplification(symbol_table).walk(rule)
+    if rule == FALSE:
+        provenance_column = str2columnstr_constant(Symbol.fresh().name)
```
demianw, line 309: maybe outsource the RHS of the equality to a function? The code of DS is already very hard to follow.

neurolang/probabilistic/tests/test_antishattering.py
```diff
@@ -0,0 +1,412 @@
+from operator import eq
```
demianw, line 1: I think we need testing for cases with more than one parameter. Something like

`P(a, x) & P(y, b) & P(z, w)` and variations of these cases.

neurolang/probabilistic/cplogic/testing.py
```diff
@@ -53,19 +53,16 @@ def build_ra_provenance_set_from_named_ra_set(nas, provenance_column):
...
-    )
+    return ProvenanceAlgebraSet(res, str2columnstr_constant(provenance_column))
 
 
 def eq_prov_relations(pas1, pas2):
     if not (
-        isinstance(pas1, ProvenanceAlgebraSet) and
-        isinstance(pas2, ProvenanceAlgebraSet)
+        isinstance(pas1, ProvenanceAlgebraSet)
+        and isinstance(pas2, ProvenanceAlgebraSet)
```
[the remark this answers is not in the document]
gzanitti: It was probably a change made by autoformatting. It is now disabled. Sorry.

PR #767, opened 2023-09-14
[REFACTOR] : Lark grammar made compatible with LALR algorithm

lalshikh:
--- Summary --------------------------------------------------------------------

A] Adaptation of the grammar
   -> files 'neurolang/frontend/datalog/standard_syntax.py'
            'neurolang/frontend/datalog/tests/test_parser.py'
            'neurolang/frontend/datalog/doc/*'

--- Details --------------------------------------------------------------------

1/ File neurolang/frontend/datalog/doc/performed_modifications.txt :

New file. Details all the modifications.

2/ Files neurolang/frontend/datalog/doc/grammar_lark.* :

Graphical version of the new grammar.

3/ Files neurolang/frontend/datalog/standard_syntax.py
     and neurolang/frontend/datalog/tests/test_parser.py :

All modifications detailed in neurolang/frontend/datalog/doc/performed_modifications.txt


demianw:
Great job @lalshikh can you please remove the file `performed_modifications.txt`? It doesn't make sense to have in the main structure. Also the duplication of the syntax on `neurolang/frontend/datalog/doc/grammar_lark.*` and the python file might prove difficult to maintain. Maybe you can move the grammar text file somewhere else and load it from the python file if you need it for the grammar plot

neurolang/frontend/datalog/standard_syntax.py
```diff
@@ -51,48 +51,47 @@
...
+?existential_body : arguments SUCH_THAT predicate ( "," predicate )*
 SUCH_THAT : "st" | ";"
 
 ?head : head_predicate
 head_predicate : identifier "(" [ arguments ] ")"
 
 ?body : conjunction
-conjunction : predicate (CONJUNCTION_SYMBOL predicate)*
-CONJUNCTION_SYMBOL : "," | "&" | "\N{LOGICAL AND}"
+conjunction : predicate ("," predicate)*
```
demianw, line 61: Can we change here the "," for the CONJUNCTION_SYMBOL or this will break the LALR(1) requirements?

neurolang/frontend/datalog/standard_syntax.py
```diff
@@ -174,7 +173,7 @@
 }
 
 
-COMPILED_GRAMMAR = Lark(GRAMMAR, debug=True)
+COMPILED_GRAMMAR = Lark(GRAMMAR, parser='lalr', debug=True)
```
demianw, line 176: do we need the "debug=True" option?

neurolang/frontend/datalog/standard_syntax.py
```diff
@@ -541,4 +533,4 @@ def _default(self, ast):
 
 def parser(code, locals=None, globals=None):
     jp = COMPILED_GRAMMAR.parse(code.strip())
-    return DatalogTransformer().transform(jp)
\ No newline at end of file
+    return DatalogTransformer().transform(jp)
```
demianw, line 536: I think that this is not the place for the datalog transformer. Parsing should not include this part of the process.

neurolang/frontend/datalog/tests/test_parser.py
```diff
@@ -1,7 +1,7 @@
 from operator import add, eq, lt, mul, pow, sub, truediv
 
 import pytest
-from lark.exceptions import UnexpectedCharacters
+from lark.exceptions import UnexpectedToken
```
demianw, line 4: I think the best would be to catch this exception inside of the parser module and catch an exception of our own, which we can include in our exceptions module. This will make the codebase more independent of the used parser.

neurolang/utils/server/tests/test_queries.py
```diff
@@ -4,7 +4,7 @@
 from uuid import uuid4
 
 import pytest
-from lark.exceptions import UnexpectedCharacters
+from lark.exceptions import UnexpectedToken
```
demianw, line 7: same comment as in neurolang/frontend/datalog/tests/test_parser.py

demianw (changes requested):
Great job! Now looking at it in more detail, some code-specific comments.

neurolang/frontend/datalog/standard_syntax.py
lalshikh: We can use CONJUNCTION_SYMBOL : "," but the previous complete expression breaks the LALR(1) requirements.

neurolang/frontend/datalog/standard_syntax.py
lalshikh: So far I need it so that the parser displays all the information I need for the interactive functionality. I'd like to keep it for now. After I finish improving the auto-completion, I will have a better idea if it's required or not.

neurolang/frontend/datalog/standard_syntax.py
lalshikh: The DatalogTransformer class evaluates the tree generated by Lark parser. It replaces the previous DatalogSemantics class used by Tatsu parser to evaluate a parser. I just left it at the same place. Should I create a new module for the evaluation of the tree part ?

neurolang/frontend/datalog/tests/test_parser.py
lalshikh: Ok. Working on it.

neurolang/frontend/datalog/standard_syntax.py
demianw: Ah OK, it was already in the tatsu version? Then my bad.

neurolang/frontend/datalog/standard_syntax.py
demianw: Ah! Great!

neurolang/frontend/datalog/standard_syntax.py
demianw: I don't understand what you mean here. You mean have just one conjunction symbol, the ","?

lalshikh: Yes.

neurolang/frontend/datalog/standard_syntax.py
lalshikh: Yes. The evaluation of the tree was already in the same file as the grammar for the Tatsu version.

PR #797, opened 2024-02-09
Next tokens facets adaptation

lalshikh:


demianw:
@lalshikh I re-ran the tests, and effectively it is due to the amount of time they take. Is there a possibility to reduce this? From the tests added, do you have the amount of time each take on your computer? At least to have relative measurements?

PR #833, opened 2026-05-18
feat: probabilistic containment and lifted-plan caching

demianw:
Add caching to probabilistic resolution to avoid redundant computation:
- `containment.py`: Cache `is_contained()` via `lru_cache` (maxsize=4096)
- `dalvi_suciu_lift.py`: Cache lifted plans with identity-safe bounded OrderedDict
- `expression_processing.py`: Remove redundant `visited` set from traversal
- `conftest.py`: Autouse fixture to clear caches between tests

Design: `(id(rule), st_key)` avoids hash collisions; thread-safe via `RLock`; LRU capped at 4096.

### DIFF
--- a/neurolang/probabilistic/dalvi_suciu_lift.py
+++ b/neurolang/probabilistic/dalvi_suciu_lift.py
@@ -1,4 +1,5 @@
 import logging
+from collections.abc import Mapping
 from functools import reduce
 from itertools import chain, combinations
 from typing import AbstractSet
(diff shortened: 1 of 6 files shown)


neurolang/probabilistic/dalvi_suciu_lift.py
```diff
@@ -114,9 +102,50 @@
     "dalvi_suciu_lift",
     "solve_succ_query",
     "solve_marg_query",
+    "clear_cache",
 ]
 
 
+# Bound cache to avoid unbounded memory growth in long-running processes.
+_MAX_DALVI_SUCU_CACHE_SIZE = 4096
```
demianw: Maybe move this to the config file ?

neurolang/probabilistic/expression_processing.py
```diff
@@ -2,8 +2,8 @@
 from typing import AbstractSet, Iterable
 
 import numpy
-from neurolang.datalog.constraints_representation import RightImplication
 
+from neurolang.datalog.constraints_representation import RightImplication
```
demianw: this should be a relative import

PR #841, opened 2026-05-22
feat: add PROB head/body syntax and aggregation to standard Datalog parser

demianw:
## Summary

Extends the standard Datalog LALR parser (`standard_syntax.py`) with probabilistic predicate syntax (`PROB`) and head aggregation, ported from the `neurolang-react` project's parser design.

## Changes

**Grammar (standard_syntax.py):**
- `prob_head_rule`: `PROB[pred] :- body` and `PROB[pred | cond] :- body`
- `prob_body_predicate`: `PROB[pred] = var` inside rule bodies
- `AGGREGATION_FUNCS` detection (count, sum, max, min, mean, avg, std) in head arguments

**Transformer (standard_syntax.py):**
- `_wrap_aggregation_args` — wraps matching function applications as `AggregationApplication`
- `__PROB__` marker system — detects PROB body predicates, strips markers, restructures into `Query` form
- Updated `head_predicate`, `query`, `rule` transformer methods

**Autocomplete fix (interactive_parsing.py):**
- Added `PROB`, `LSQB`, `RSQB` to `TERMINALS_TO_CATEGORIES` mapping

**Tests:**
- 8 new tests covering aggregation, PROB head rule, and PROB body predicate
- 510 tests pass (no regressions)

## Verification

```
510 passed, 12 skipped, 1 xpassed in 13.52s
```


neurolang/frontend/datalog/tests/test_parser.py
```diff
@@ -476,3 +476,279 @@ def test_command_syntax():
...
+def test_prob_body_simple():
+    """ans(x, p):-PROB[pred(x)]=p → Query extracting pred and p."""
+    res = parser("ans(x, p):-PROB[pred(x)]=p")
+    pred = Symbol("pred")
+    x = Symbol("x")
+
+    q = res.formulas[0]
+    assert isinstance(q, Query)
+    assert q.head.functor == Symbol("ans")
+    body_formulas = q.body.formulas if hasattr(q.body, 'formulas') else (q.body,)
```
demianw: can you make this test check more explicitely the IR generated? Like in https://github.com/NeuroLang/NeuroLang/pull/841/changes#diff-e963eb79a31ab85d62e736389d0db0857cb7a28992d99e1abed92cc48e52c54bR581

neurolang/frontend/datalog/tests/test_parser.py
```diff
@@ -476,3 +476,279 @@ def test_command_syntax():
...
+    res = parser("ans(x, p):-filter(x) & PROB[pred(x)]=p")
+
+    q = res.formulas[0]
+    assert isinstance(q, Query)
+    assert q.head.functor == Symbol("ans")
+    body = q.body
+    assert isinstance(body, Conjunction)
+    body_atoms = list(body.formulas)
+    assert len(body_atoms) == 2
+    functors = {
```
demianw: Again check more explicitely the IR generated like in n you make this test check more explicitely the IR generated? Like in https://github.com/NeuroLang/NeuroLang/pull/841/changes#diff-e963eb79a31ab85d62e736389d0db0857cb7a28992d99e1abed92cc48e52c54bR581 and generalize that to all similar tests in this file

neurolang/frontend/datalog/tests/test_parser.py
demianw: Addressed in commit 8574ef1d. Added explicit IR assertions (DatalogProgram structure checks) to the aggregation and PROB syntax tests — they now verify the actual program structure (rules, heads, bodies, literals) rather than just roundtrip parsing.

neurolang/frontend/datalog/tests/test_parser.py
demianw: Addressed in commit 8574ef1d. All new syntax tests (aggregation in head/query, PROB head rules, PROB body predicates, MARG/SUCC predicates) now include explicit IR assertions — they construct the expected DatalogProgram/Union with Symbol, Implication, Query, Conjunction, Condition, AggregationApplication, etc. and assert equality, following the pattern from the referenced test.

neurolang/frontend/datalog/tests/test_parser.py
```diff
@@ -476,3 +476,293 @@ def test_command_syntax():
...
+
+# ── New syntax: MARG body predicate ──────────────────────────────────────────
+
+def test_marg_body_simple():
+    """ans(x, p):-MARG[pred(x)]=p → same IR as PROB body predicate."""
+    ans = Symbol("ans")
+    pred = Symbol("pred")
+    x = Symbol("x")
+    p = Symbol("p")
+    res = parser("ans(x, p):-MARG[pred(x)]=p")
```
demianw, line 701: This test is wrong, the resulting code should be a representation of:

fresh[x, PROB(x)] = pred(x)
ans[x, p] = fresh(x, p)

in terms of the IR

propagate this to all other similar tests and then fix the code such that it passes.

demianw: Fixed in commit 14e83d77. `_build_prob_rule` now desugars `PROB[pred(x)]=p` into two formulas:

1. `Implication(fresh(x, PROB(x)), pred(x))` — the fresh predicate rule
2. `Query(ans(x, p), fresh(x, p))` — the main query using the fresh predicate

All 5 PROB/MARG body predicate tests were rewritten with structural IR assertions (`isinstance`, fresh name prefix check, functor identity across formulas) that verify this exact structure without relying on `Symbol.fresh()` counter values.

neurolang/frontend/datalog/tests/test_parser.py
demianw: Confirmed in commit 14e83d77. The aggregation and PROB head/tests already had explicit IR assertions. The PROB/MARG body tests have now been updated with the correct Implication+Query desugaring structure.

neurolang/frontend/datalog/tests/test_parser.py
```diff
@@ -590,36 +590,73 @@ def test_prob_head_rule_conditional():
...
+    assert isinstance(res, Union)
+    assert len(res.formulas) == 2
+
+    # Formula 0: Implication fresh(x, PROB(x)) :- pred(x)
+    impl = res.formulas[0]
+    assert isinstance(impl, Implication)
+    fresh_head = impl.consequent
+    assert isinstance(fresh_head, FunctionApplication)
+    assert isinstance(fresh_head.functor, Symbol)
+    assert fresh_head.functor.name.startswith("fresh_")
```
demianw: I think you can check this with symbol.is_fresh, propagate this to this whole test file

neurolang/frontend/datalog/tests/test_parser.py
```diff
@@ -590,36 +590,73 @@ def test_prob_head_rule_conditional():
...
-            ans(x, p),
-            Conjunction((pred(x),)),
-        ),
-    ))
-    assert res == expected
+
+    assert isinstance(res, Union)
+    assert len(res.formulas) == 2
+
+    # Formula 0: Implication fresh(x, PROB(x)) :- pred(x)
```
demianw: This way to check is  too verbose and not very readable. In other tests there are ways to check that two formulas are equivalent. If not do it with a helper function that checks that they are equivalent. Propagate to the whole file.

neurolang/frontend/datalog/tests/test_parser.py
demianw: Done in commit e226d304. Replaced all `fresh_head.functor.name.startswith("fresh_")` with `fresh.is_fresh` and propagated to all 5 PROB/MARG body tests.

neurolang/frontend/datalog/tests/test_parser.py
demianw: Done in commit e226d304. Rewrote all 5 tests to extract the fresh symbol from the result, assert is_fresh, then compare with `==` against an expected Union built from that symbol. This matches the existing pattern used for `...` wildcard tests (line 72-77) and is much more compact: 71 insertions / 139 deletions.

demianw: Coverage update (commit 9faf1c5c): the conditional PROB body path (`PROB[pred(x) // cond(x)]=p`) was also untested and had a bug — the `//` token leaked into output. Both are now fixed and covered by `test_prob_body_conditional` and `test_prob_body_conditional_with_filter`.

Remaining uncovered lines are defensive/dead code: the `else` branch in `_build_prob_rule` (pred would need to be neither FunctionApplication nor Conjunction, impossible per grammar), `result_var is None` paths (grammar requires `=argument`), and `prob_body_predicate` `len(ast) == 3` (no grammar path produces exactly 3 elements).

PR #845, opened 2026-05-26
feat: contains/Destroy Datalog operator with tuple literal grammar

demianw:
Adds a `contains(R, (I, J, K))` Datalog operator that decomposes a brain region into individual voxel coordinates.

This PR contains only the destroy-specific changes. The engine registry infrastructure (YAML engine definitions, templates, CLI) is in PR #840.

Changes:
- `tuple_literal` grammar rule in `standard_syntax.py` for `(a, b, c)` syntax
- `ExplicitVBR.__contains__` for voxel membership checks
- `contains` builtin registered in `QueryBuilderDatalog` frontend
- `PandasSet.explode` handles Region columns by converting voxel arrays to iterable tuples

Testing: region tests (19), frontend tests (34), RA tests (38) pass.

### DIFF
--- a/neurolang/frontend/datalog/standard_syntax.py
+++ b/neurolang/frontend/datalog/standard_syntax.py
@@ -131,7 +131,9 @@ fact : constant_predicate
 constant_predicate : identifier "(" (literal | ext_identifier) ("," (literal | ext_identifier))* ")"
                    | identifier "(" ")"
 
-?literal : number | text
+?literal : number | text | tuple_literal
+
+tuple_literal : "(" argument ("," argument)+ ")"
 
 ext_identifier : "@" identifier
 identifier : cmd_identifier | identifier_regexp
@@ -526,6 +528,15 @@ class DatalogTransformer(Transformer):
 
     def text(self, ast):
         return Constant((ast[0].replace("'", "")).replace('"', ''))
+
+    def tuple_literal(self, ast):
+        from neurolang.expressions import Constant as ConstantExpr
+        values = tuple(a for a in ast if not isinstance(a, str))
+        result = ConstantExpr(values)
+        for v in values:
+            if isinstance(v, Symbol):
+                result._symbols.add(v)
+        return result
 
     def pos_int(self, ast):
         return Constant(int((ast[0]).value))
--- a/neurolang/frontend/datalog/standard_syntax.py
+++ b/neurolang/frontend/datalog/standard_syntax.py
@@ -1,5 +1,6 @@
 import json
 import os
+from typing import Tuple
 
 from lark import Lark, Transformer
 from lark.exceptions import UnexpectedToken, UnexpectedCharacters, LarkError
@@ -24,6 +25,7 @@ from ...probabilistic.expressions import (
     Condition,
     ProbabilisticFact
 )
+from ...type_system import Unknown
 from ...utils.interactive_parsing import LarkCompleter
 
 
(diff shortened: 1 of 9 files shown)


neurolang/frontend/datalog/standard_syntax.py
```diff
@@ -527,6 +529,15 @@ def cmd_identifier(self, ast):
     def text(self, ast):
         return Constant((ast[0].replace("'", "")).replace('"', ''))
 
+    def tuple_literal(self, ast):
+        from neurolang.expressions import Constant as ConstantExpr
+        values = tuple(a for a in ast if not isinstance(a, str))
+        result = ConstantExpr(values)
```
demianw: the parsing is faster if the type of the expression is specified at construction. Also why ConstantExpr and not Constant?

neurolang/frontend/query_resolution_datalog.py
```diff
@@ -5,6 +5,7 @@
 as well as Region and Neurosynth capabilities
 """
 from collections import defaultdict
+import operator
 from neurolang.datalog.magic_sets import magic_rewrite
 from neurolang.exceptions import (
```
demianw: these two imports should be relative

neurolang/frontend/query_resolution_datalog.py
```diff
@@ -105,6 +106,7 @@ def __init__(
         )
         self.translate_expression_to_datalog = TranslateToDatalogSemantics()
         self.datalog_parser = datalog_parser
+        self.add_symbol(operator.contains, name="contains")
```
demianw: Isn't there a part of the code where all operators are added together automatically ?

neurolang/utils/relational_algebra_set/pandas.py
```diff
@@ -720,23 +720,37 @@ def explode(self, src_column, dst_columns):
         if self._container is None:
             return self.copy()
 
+        # Convert column values with a voxels attribute (e.g. ExplicitVBR)
+        # into iterable tuples so pandas .explode() can flatten them.
+        container = self._container
+        if len(container) > 0:
+            col = container[src_column]
+            first = col.iloc[0]
+            if hasattr(first, 'voxels') and not isinstance(first, (str, bytes)):
```
demianw: You shouldn't hardcode the voxels container at this level. If this needs to be addressed it should be at the Sugar level in the frontend.

neurolang/utils/relational_algebra_set/pandas.py
```diff
@@ -720,23 +720,37 @@ def explode(self, src_column, dst_columns):
...
+                container = container.copy()
+                container[src_column] = col.apply(
```
demianw: This should be avoided as ussing apply versus the pandas.dataframe.explode method will be much slower

neurolang/frontend/datalog/sugar/spatial.py
```diff
@@ -257,6 +257,87 @@ def solve_spatial_bound(
...
+
+    @add_match(
+        Implication,
+        lambda imp: (
+            isinstance(imp.antecedent, Conjunction)
+            and any(
+                isinstance(a.functor, Constant)
+                and a.functor.value is operator.contains
+                for a in extract_logic_atoms(imp.antecedent)
+            )
```
demianw: The guard sseems too relaxed, it should also verify that the relation object with the column with the "voxels" method is there

neurolang/frontend/datalog/sugar/spatial.py
```diff
@@ -257,6 +257,87 @@ def solve_spatial_bound(
...
+                isinstance(atom.functor, Constant)
+                and atom.functor.value is operator.contains
+            ):
+                continue
+            if len(atom.args) < 1:
+                continue
+            region_var = atom.args[0]
+            if not isinstance(region_var, Symbol):
+                continue
+            for binding_atom in atoms:
```
demianw: the code inside of this loop seems overly complex. Can you simplify it by adding a stricter guard and by refactoring ?

neurolang/frontend/datalog/standard_syntax.py
```diff
@@ -527,6 +529,14 @@ def cmd_identifier(self, ast):
     def text(self, ast):
         return Constant((ast[0].replace("'", "")).replace('"', ''))
 
+    def tuple_literal(self, ast):
+        values = tuple(a for a in ast if not isinstance(a, str))
```
demianw, line 538: seems that if a is a string, this will end in a bug. Can you double check ? Also if you iterate like this to get the values, you should be also able to get the types and then to set up the type of the constant or add an explicit unknown.

neurolang/frontend/datalog/standard_syntax.py
```diff
@@ -527,6 +529,14 @@ def cmd_identifier(self, ast):
...
+        result = Constant(values)
```
demianw: add the type even if it is [Tuple[Uknown, .........]] with as many unknowns as elements there are.

demianw:
Solved by PR #850

PR #850, opened 2026-05-30
feat: Add syntactic sugar for ExplicitVBR destroy with tuple-destructured voxel coordinates

demianw:
## Summary
This PR extends the Datalog frontend with syntactic sugar for exploding ExplicitVBR region columns into individual voxels via the contains/Destroy pipeline.

## Changes
- **FrozenNDArray**: New immutable, hashable np.ndarray subclass for fast voxel handling without Python tuple conversion overhead.
- **MRO ordering**: Moved TranslateRegionDestroy AFTER TypeResolutionMixin so type resolution runs before the sugar guard.
- **Type guard**: Added is_leq_informative(a.args[0].type, ExplicitVBR) to _is_contains_atom to prevent false positives on non-region contains patterns.
- **Tuple-destructured voxels**: Fixed _build_tuple_constant to include inner Symbols in _symbols, enabling rules like:
  ```datalog
  res(l, x, y, z) :- region(l, r) \u0026 contains(r, (x, y, z))
  ```
- **Tests**: Added 5 tests covering set destroy, region destroy, non-region ignored, multi-column, and tuple-destructured coordinates.

## Test Results
All 318 related tests pass (10 skipped).


neurolang/frontend/tests/test_frontend.py
```diff
@@ -798,6 +798,110 @@ def test_neurolang_dl_set_destroy():
...
+    assert len(ans) == 3
+    expected = {
+        ((1, 0, 0),),
+        ((1, 1, 0),),
+        ((2, 0, 0),),
+    }
+    assert {tuple(tuple(v) for v in row) for row in ans} == expected
+
+
+def test_neurolang_dl_region_destroy_non_region_ignored():
```
demianw, line 829: Also add a test where the test set is :

numbers = neurolang.add_tuple_set(
        [("a", (1, 2, 3))], name="numbers"
    )

which should be destroyed into 
{("a", 1), ("a", 2), ("a", 3)}

If it doesn't happen, fix it.

demianw: Added `test_neurolang_dl_region_destroy_tuple_edb_column` verifying this exact pattern:

```python
numbers = neurolang.add_tuple_set([("a", (1, 2, 3))], name="numbers")
e.ans[e.l, e.v] = numbers[e.l, e.t] & contains_(e.t, e.v)
# Result: {('a', 1), ('a', 2), ('a', 3)}
```

It passes — the existing Destroy RA operator already handles exploding tuple-valued EDB columns via `pandas.explode()`. No code changes needed.

PR #854, opened 2026-06-05
fix(squall): resolve 'the X' anaphora across 'given' boundary in conditional probability queries

demianw:
## Problem

In SQUALL conditional probability queries like:
```
every Voxel that a SelectedStudy reports given the SelectedStudy mentions 'emotion'
```
`a SelectedStudy` and `the SelectedStudy` were producing different existential variables (s₃ vs s₄), causing incorrect query semantics. The root cause: Lark evaluates the conditioning side (`s = \"the SelectedStudy mentions 'emotion'\"`) bottom-up before the `det1_some` handler creates the existential from `a SelectedStudy` inside the relative clause.

### New expression class: `AnaphoraPredicate`
\texttt{neurolang/logic/\_\_init\_\_.py} — extends `ExistentialPredicate` with a `noun_name` Symbol child. Marks cross-boundary `the X` references.

### New walker: `AnaphoraResolutionWalker`
\texttt{neurolang/frontend/datalog/anaphora\_resolution.py} — post-processing `ExpressionWalker` that:
1. Collects `AnaphoraPredicate` markers from the tree (before walking children)
2. Searches both sides of `Condition` for matching noun predicates
3. Unifies existential variables via `ReplaceSymbolWalker`
4. Strips remaining markers to plain `ExistentialPredicate`

### Transformer integration
\texttt{neurolang/frontend/datalog/squall\_syntax\_lark.py} — modified `det_the` to produce `AnaphoraPredicate` when a `noun_name` is available; integrated `AnaphoraResolutionWalker` into `parser()` as a post-simplification step.

## Testing
- 2 new tests covering `AnaphoraPredicate` class construction and integration with SQUALL parser
- Full test suite: **1388 passed, 17 skipped, 0 failures** — no regressions

## Notes
- `AnaphoraResolutionWalker` follows NeuroLang's IR processing convention: extends `ExpressionWalker` with `@add_match` handlers, uses guard predicates to avoid interfering with standard walker behavior.
- Only fires when `AnaphoraPredicate` nodes are present — zero overhead for queries without anaphora.

### DIFF
--- a/neurolang/frontend/datalog/tests/test_squall_parser.py
+++ b/neurolang/frontend/datalog/tests/test_squall_parser.py
@@ -96,28 +96,21 @@ class LogicWeakEquivalence(ExpressionWalker):
     def eq_expression(self, expression):
         left, right = expression.args
         if type(left) is not type(right):
-            # Unknown/Any type parameter unification:
-            # same generic root + unifiable types → compare by structure
-            left_meta = type(left)
-            right_meta = type(right)
+            # Unify Unknown/Any across parameterized types
+            lt = type(left)
+            rt = type(right)
             if not (
-                isinstance(left_meta, ParametricTypeClassMeta) and
-                isinstance(right_meta, ParametricTypeClassMeta)
+                isinstance(lt, ParametricTypeClassMeta) and
+                isinstance(rt, ParametricTypeClassMeta)
             ):
                 return False
-            # __generic_class__ only exists on parameterized subclasses;
-            # the base class itself is its own generic root
-            left_root = getattr(left_meta, '__generic_class__', None) or left_meta
-            right_root = getattr(right_meta, '__generic_class__', None) or right_meta
+            left_root = getattr(lt, '__generic_class__', lt)
+            right_root = getattr(rt, '__generic_class__', rt)
             if not (
                 left_root is right_root and
                 (
-                    is_leq_informative(
-                        left_meta._nl_class_type, right_meta._nl_class_type
-                    ) or
-                    is_leq_informative(
-                        right_meta._nl_class_type, left_meta._nl_class_type
-                    )
+                    is_leq_informative(lt._nl_class_type, rt._nl_class_type) or
+                    is_leq_informative(rt._nl_class_type, lt._nl_class_type)
                 )
             ):
                 return False
@@ -147,16 +140,16 @@ class ConditionAwareEqMixin(PatternWalker):
     def eq_logic_operator(self, expression):
         left, right = expression.args
         results = []
-        for l, r in zip(left.unapply(), right.unapply()):
-            if isinstance(l, tuple) and isinstance(r, tuple):
-                if len(l) != len(r):
+        for lv, rv in zip(left.unapply(), right.unapply()):
+            if isinstance(lv, tuple) and isinstance(rv, tuple):
+                if len(lv) != len(rv):
                     return False
                 results.append(all(
-                    self.walk(EQ(ll, rr))
-                    for ll, rr in zip(l, r)
+                    self.walk(EQ(lvv, rvv))
+                    for lvv, rvv in zip(lv, rv)
                 ))
             else:
-                results.append(self.walk(EQ(l, r)))
+                results.append(self.walk(EQ(lv, rv)))
         return all(results)
 
 
--- a/neurolang/frontend/datalog/tests/test_squall_parser.py
+++ b/neurolang/frontend/datalog/tests/test_squall_parser.py
@@ -11,7 +11,10 @@ from ....expression_pattern_matching import add_match
 from ....expression_walker import (
     ExpressionWalker, PatternWalker, ReplaceExpressionWalker
 )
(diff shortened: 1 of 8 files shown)


neurolang/frontend/datalog/tests/test_squall_parser.py
```diff
@@ -97,6 +98,55 @@ def weak_logic_eq(left, right):
     return LogicWeakEquivalence().walk(EQ(left, right))
 
 
+class ConditionAwareEqMixin(ExpressionWalker):
```
demianw: If it's a Mixin it should inherit from PatternWalker, can you correct this?

neurolang/logic/__init__.py
```diff
@@ -149,6 +149,30 @@ def __repr__(self):
         return r
 
 
+class AnaphoraPredicate(ExistentialPredicate):
```
demianw: this type of predicate is specific to SQUALL not a pure logic operator so it should be moved to the frontend.datalog package

neurolang/utils/cli.py
```diff
@@ -321,150 +332,201 @@ def _execute_squall_program(
     return nl.execute_squall_program(program_text)
 
 
-def _format_ir(expr, fresh_map=None, _counter=None):
-    """Format a NeuroLang IR expression as human-readable Datalog-like text.
+class DatalogPrettyPrinter(PatternWalker):
```
demianw: this class should be moved outside of this file. Probably to the neurolang.frontend.datalog package or to the utils package.

demianw:
Addressed all review comments:

1. **ConditionAwareEqMixin → PatternWalker**: Changed base class from `ExpressionWalker` to `PatternWalker` (`neurolang/frontend/datalog/tests/test_squall_parser.py`).

2. **AnaphoraPredicate → frontend.datalog**: Moved from `neurolog/logic/__init__.py` to `neurolang/frontend/datalog/anaphora_resolution.py`. Updated all imports (squall_syntax_lark.py, test_squall_parser.py). The `AnaphoraPredicate` is now in the SQUALL frontend package alongside the resolution walker.

3. **DatalogPrettyPrinter → dedicated module**: Extracted from `neurolang/utils/cli.py` into a new `neurolang/frontend/datalog/pretty_printer.py`. The `cli.py` now imports the class from the new module. All imports updated in `test_cli.py`.

Also rebased onto upstream/master (resolved merge conflicts with PR #853).

neurolang/frontend/datalog/tests/test_squall_parser.py
```diff
@@ -97,6 +100,55 @@ def weak_logic_eq(left, right):
...
+                    return False
+                results.append(all(
+                    self.walk(EQ(ll, rr))
+                    for ll, rr in zip(l, r)
+                ))
+            else:
+                results.append(self.walk(EQ(l, r)))
+        return all(results)
+
+    @add_match(EQ(..., ...))
```
demianw: why is this one needed ? Doesn't it take from LogicWeakEquivalence when you combine them?

demianw: It is needed because `LogicWeakEquivalence.eq_expression` (the base) uses `expression.args[0] == expression.args[1]` — plain Python `==`. This breaks for `FunctionApplication` nodes where type parameters differ (e.g. `FunctionApplication[Unknown]` vs `FunctionApplication[Any]`). The asymmetry is in `Expression.__eq__`, not in this file.

The mixin overrides with `left.unapply() != right.unapply()` which compares the structural tuple (constructor, args, type) and avoids the `__eq__` asymmetry entirely.

Without this override, `ConditionAwareLogicWeakEquivalence` would fall through to the base `eq_expression` and fail on expressions with mismatched type parameters — the very thing the `unapply()` trick exists to solve.

demianw: Actually, let me investigate this more carefully. You may be right — if `Any` and `Unknown` should unify in the type system, then fixing it at the source would be better than papering over it with `unapply()` in the test helper. Let me look at the type comparison logic to see if this is a bug or by design.

demianw: Took a different approach: instead of the unapply() override in the mixin, I fixed `LogicWeakEquivalence.eq_expression` itself.

The root cause is that `Expression.__eq__` does `isinstance(other, type(self))`, which fails when `FunctionApplication[Unknown]` (parser output) is compared to `FunctionApplication[Any]` (test construction) — these are different classes created by the parameterized metaclass.

The fix: `eq_expression` compares by `unapply()` structural tuple directly, with a type-compatibility gate that resolves `Any`/`Unknown` mismatches via `ParametricTypeClassMeta` introspection + `is_leq_informative`. The `__generic_class__` quirk (only exists on parameterized subclasses, not on the base class itself) is handled by falling back to `left_meta` / `right_meta` when `__generic_class__` is absent.

The `eq_expression` override in `ConditionAwareEqMixin` is now removed since the base does the right thing.

Pushed in c90d6c95.

PR #864, opened 2026-06-13
fix(squall): allow dimension keywords (Probability, Value) as nouns in compound quantifiers

demianw:
## Problem

`Probability` and `Value` are `DIM_PROBABILITY` / `DIM_VALUE` reserved terminals, excluded from the `UPPER_NAME` regex by negative lookahead. This means they cannot be parsed as regular nouns in SQUALL phrases like:

```
define as activation_probability for every Voxel in 3D
  and for every Probability
  where a Schaefer_label labels the Voxel
  and Label_reports the Probability.
```

The parser fails at "Probability" because no terminal matches `P`.

### 1. Grammar (neurolang_natural.lark)
Add `dimension_noun` as an alternative in `noun1` so dimension keywords can serve as regular nouns:

```diff
 noun1 : intransitive
+     | dimension_noun
```

### 2. Flatten tuple variables in compound quantifier heads (squall_syntax_lark.py)
`Voxel in 3D` produces a tuple of 3 fresh symbols `(s0, s1, s2)` as the quantifier variable. Without flattening, `verb(*head_vars)` produces `activation_probability((s0, s1, s2), s3)` instead of the correct flat `activation_probability(s0, s1, s2, s3)`.

The fix flattens tuple variables in `rule_body2_where` before appending to `head_vars`.

## Result

The SQUALL rule now correctly parses to Datalog:

```
activation_probability(s₀, s₁, s₂, s₃) :-
  voxel(s₀, s₁, s₂),
  probability(s₃),
  schaefer_label(s₄),
  labels(s₄, s₀, s₁, s₂),
  label_reports(s₄, s₃).
```

## Testing
- All 283 existing tests pass
- New test `test_dimension_noun_in_compound_quantifier` covers the exact rule above, verifying head arity, predicate arities, and anaphora variable sharing

### DIFF
--- a/neurolang/frontend/datalog/neurolang_natural.lark
+++ b/neurolang/frontend/datalog/neurolang_natural.lark
@@ -187,8 +187,7 @@ adj1 : intransitive
 adjn : transitive_multiple
 
 noun1 : intransitive
-      | DIM_PROBABILITY
-      | DIM_VALUE
+     | dimension_noun
 noun2 : transitive
 
 verb1 : intransitive
--- a/neurolang/frontend/datalog/neurolang_natural.lark
+++ b/neurolang/frontend/datalog/neurolang_natural.lark
@@ -187,7 +187,8 @@ adj1 : intransitive
 adjn : transitive_multiple
 
 noun1 : intransitive
-     | dimension_noun
+     | DIM_PROBABILITY
+     | DIM_VALUE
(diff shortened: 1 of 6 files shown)


demianw:
Replaced by #865 — this PR had stale polars-branch history that couldn't be cleanly updated without force push

